PR #1, opened 2017-04-07
Lambda spike

rupertbates:
I'm opening a PR for this so that people can see what I've been up to and comment on it or play around with it.

### What's in the spike:
* A CloudFormation template which creates
    * An AWS step functions state machine
    * 3 Lambdas which make up 3 states of the state machine
* Classes which implement the Lambdas and pass around model objects as state whilst hiding the complexity of working with AWS input and output streams.

To execute the state machine, you will have to run it from the AWS console for now, passing in json similar to the following:
    `{
    "id": "123",
  	"displayName": "blah"
    }`

To build and deploy it you can use deploy.sh which will build a jar and push it to S3 (you'll need valid membership creds) and you can update the step function from the AWS console using the cloudformation template. It seems that sometimes you need to delete and recreate the step function rather than just updating in order to get it to pick up code changes.

### Error handling
I have also spiked out the error handling strategy we can use to cope with temporary outages in 3rd party system:

The last lambda in the Step function is called FlakeyService and will always fail the first time it is called by throwing an exception of type NonFatalException. It will then succeed on subsequent invocations.

The Monthly Contributions step function is configured to retry up to 2 times when FlakeyService fails with a NonFatalException using the config below:

```
        "FlakeyService": {
                  "Type": "Task",
                  "Resource": "${flakeyServiceArn}",
                  "Retry": [ {
                    "ErrorEquals": ["com.gu.support.workers.exceptions.NonFatalException"],
                    "IntervalSeconds": 1,
                    "MaxAttempts": 2,
                    "BackoffRate": 2.0
                           } ],
                  "End": true
                }
```
Since FlakeyService always succeeds on the second attempt the step function will retry and then succeed. If we wanted to fail completely for instance due to a serious logic error then we could throw a FatalException (or in fact any exception we are not explicitly retrying on) and the whole step function would fail.


rupertbates:
Merging this so we can take it forward as a team

PR #2, opened 2017-04-21
Project reorganisation

rupertbates:
This is a reorganisation of the project to include a common module which can be used by all of the individual worker projects.

New structure:

support-workers
-- common
-- monthly-contributions
-- [other-worker-project]
...

It also contains some okhttp stuff which I've hauled in from membership-common and the start of work on the CreatePaymentMethod lambda, all of which I started before I decided that it would be better to get the project structure right first.


common/src/main/scala/com.gu.okhttp/package.scala
```diff
@@ -0,0 +1,34 @@
...
+  implicit class RickOkHttpClient(client: OkHttpClient) extends LazyLogging {
+
+    def execute(request: Request): Future[Response] = {
+      val p = Promise[Response]()
+
+      client.newCall(request).enqueue(new Callback {
+        override def onFailure(call: Call, e: IOException) {
+          val sanitizedUrl = s"${request.url().uri().getHost}${request.url().uri().getPath}" // don't log query string
+
+          if (!(sanitizedUrl == "d1s44y2vc5rmbm.cloudfront.net/v3/users/me/owned_events/")) // FIXME: Temporarily disable logging of CloudFront errors to prevent log spamming as we are getting 30000 timeouts per day
```
davidfurey: Do we still need this?

common/src/main/scala/com/gu/support/workers/helpers/Timing.scala
```diff
@@ -0,0 +1,27 @@
...
+import com.gu.support.workers.monitoring.CloudWatch
+import com.typesafe.scalalogging.LazyLogging
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+
+object Timing extends LazyLogging {
+
+  def record[T](cloudWatch: CloudWatch, metricName: String)(block: => Future[T]): Future[T] = {
```
davidfurey: I've been talking to Roberto about potentially removing these Cloudwatch metrics from membership-frontend, since we pay for them but don't seem to use them.  I think we should leave them out of support-worker unless you think we will make use of them.

common/src/main/scala/com.gu.okhttp/package.scala
rupertbates: Who know? I'm going to remove it because we don't really want to start of with unexplained hacks in place :-)

common/src/main/scala/com/gu/support/workers/helpers/Timing.scala
rupertbates: Nope, happy to do that, it simplifies the code.

PR #4, opened 2017-04-24
Config

rupertbates:
This PR puts some basic config infrastructure in place for the support-workers project

From the readme:

Config for the app is split into public values in [application.conf](common/src/main/resources/application.conf)
and private values which are loaded from S3 when the lambdas start up. 

When running locally you can avoid loading config from S3 on each run and load it from a local file instead as follows:

* Retrieve credentials from S3 with the command:

    `sudo aws s3 cp s3://support-workers-private/DEV/support.private.conf /etc/gu/support.private.conf --profile membership`

* Set an environment variable to false by adding the following to your bashrc:

    `export GU_SUPPORT_WORKERS_LOAD_S3_CONFIG=false`


common/src/test/scala/com.gu.config/ConfigSpec.scala
```diff
@@ -0,0 +1,13 @@
+package com.gu.config
+import com.typesafe.scalalogging.LazyLogging
+import org.scalatest.{FlatSpec, Matchers}
+
+class ConfigSpec extends FlatSpec with Matchers with LazyLogging {
```
svillafe: Tests 😄

PR #5, opened 2017-04-25
Create payment method lambda

rupertbates:
This PR creates a working implementation of the 'Create payment method' step of the monthly checkout flow. 

It expects to receive json matching the current json which is generated by the checkout form in membership-frontend:
### Stripe:
    {
      "userId": "12345",
      "stripeToken": "$stripeToken"
    }
### PayPal 
    {
       "paypalBaid": "$validBaid"
     }

And if all goes well will return a CreditCardReferenceTransaction (Stripe) or PayPalReferenceTransaction. For Stripe customers it will also create a customer via the Stripe api and so is not idempotent.

The PR itself is quite large because it includes new classes for handling web requests (using Okhttp) as well as  DTO/model classes for interacting with the Stripe and PayPal services. Also I optimised imports across the whole project which changed quite a lot of files...

### TODOs
* There is currently no mechanism in place for using different config for test users, this is something which will be developed separately and should work transparently for all the lambdas.
* Error handling is not that sophisticated at the moment, we will need to differentiate between temporary failures which should be retried and permanent failures which should abort the entire flow.


common/src/main/scala/com/gu/i18n/CountryGroup.scala
```diff
@@ -0,0 +1,315 @@
...
+    Country("KI", "Kiribati"),
+    Country("NR", "Nauru"),
+    Country("NF", "Norfolk Island"),
+    Country("TV", "Tuvalu")
+  ), AUD, PostCode)
+
+  val NewZealand = CountryGroup("New Zealand", "nz", Some(C.NewZealand), List(
+    C.NewZealand,
+    Country("CK", "Cook Islands")
+  ), NZD, PostCode)
```
Ap0c: Why do we still treat New Zealand differently from RoW? If we were to put it under `INT` would it make Zuora/Salesforce unhappy?

@rtyley @paulbrown1982

common/src/main/scala/com/gu/i18n/CountryGroup.scala
```diff
@@ -0,0 +1,315 @@
...
+  )
+
+  val countries: List[Country] = allGroups.flatMap(_.countries).sortBy(_.name)
+
+  def countryByCode(str: String): Option[Country] = countries.find { _.alpha2 == str }
+
+  def countryByName(str: String): Option[Country] = countries.find { _.name == str }
+
+  // This is because there was an inconsistency in the code where we were writing a country name
+  // in Identity but then trying to find it by code. It's not clear anymore which we have in our systems; probably both
```
Ap0c: 🙃

common/src/main/scala/com/gu/i18n/Currency.scala
```diff
@@ -0,0 +1,44 @@
...
+  }
+  case object EUR extends Currency {
+    override def glyph = "€"
+    override def iso = "EUR"
+  }
+  case object NZD extends Currency {
+    override def glyph = "$"
+    override def prefix = Some("NZ")
+    override def iso = "NZD"
+  }
```
Ap0c: Similar question as before, do we support NZD? If we were to remove it would other systems become unhappy?

@rtyley @paulbrown1982

PR #9, opened 2017-05-10
Add SendThankYouEmail lambda

davidfurey:
Puts message on SQS queue to send thank-you email


jacobwinch:
Looks good to me 👍

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
```diff
@@ -0,0 +1,33 @@
...
+    logger.info(s"state: $state")
+    sendEmail(state)
+  }
+
+  def sendEmail(state: CreateZuoraSubscriptionState): Future[Unit] = {
+    thankYouEmailService.send(ThankYouFields(
+      email = state.user.primaryEmailAddress,
+      created = DateTime.now(),
+      amount = state.amount,
+      currency = "GBP",
```
rupertbates: I think it would be good to pass this through in the State objects so we can add new ones more easily in future. We can make a separate PR for that though as it is used in the Zuora lambda as well.

cloud-formation/src/state-machine.yaml
```diff
@@ -12,6 +12,10 @@ States:
   CreateZuoraSubscription:
     Type: Task
     Resource: "$${CreateZuoraSubscriptionLambda.Arn}"
+    Next: SendThankYouEmail
+  SendThankYouEmail:
+    Type: Task
+    Resource: "$${SendThankYouEmailLambda.Arn}"
     Next: FlakeyService
```
rupertbates: We should probably get rid of the FlakeyService lambda now! Happy to do that in another PR though

PR #10, opened 2017-05-11
Create Zuora Subscription lambda

rupertbates:
## Create Zuora Subscription
This PR adds the functionality to create Zuora subscription to the monthly contributions check out step function. 

There are a lot of changes here but quite a few of them are down to Scalastyle or a refactor of the WebserviceHelper class to make it more flexible by allowing different requests to have different return types.

One area that is worth pointing out is the CapitalizationEncoder class and the CustomCodecs class which uses it. These allow us to map between the decapitalised field names of the case classes in our model and the capitalised versions which the Zuora api expects. It may well be that there is a simpler way of doing this but I felt like I'd spent enough time on it up till this point and needed to move on.

### Notes:
* This PR requires new config in the private conf file
* The output state from the Zuora step is currently just a string. As the next step is currently being redeveloped I have left it at that rather than introducing a new state object which will probably be changed.


common/src/main/scala/com/gu/zuora/encoding/CapitalizationEncoder.scala
```diff
@@ -0,0 +1,34 @@
...
+      _.withFocus(
+        _.mapObject(decapitalizeFields)
+      )
+    )
+
+  def capitalizingEncoder[A](implicit encode: Lazy[DerivedObjectEncoder[A]]): ObjectEncoder[A] =
+    deriveEncoder[A].mapJsonObject(capitalizeFields)
+
+  def modifyFields(json: JsonObject)(f: String => String): JsonObject = {
+    val newFields = json.fields.map(str => f(str)).zip(json.values)
```
davidfurey: Would this be more or less clear?

 `json.toVector.map { case (k, v) => f(k) -> v }`

common/src/main/scala/com/gu/zuora/encoding/CustomCodecs.scala
```diff
@@ -0,0 +1,57 @@
...
+trait CustomCodecs {
+  //Request encoders
+  implicit val encodeCurrency: Encoder[Currency] = Encoder.encodeString.contramap[Currency](_.iso)
+  //Encodes a Currency as its iso code eg. {"Currency": "GBP"}
+  implicit val encodePaymentGateway: Encoder[PaymentGateway] = Encoder.encodeString.contramap[PaymentGateway](_.name)
+  implicit val encodeCountryAsAlpha2: Encoder[Country] = Encoder.encodeString.contramap[Country](_.alpha2)
+
+  //Account object encoding - unfortunately the custom field name of the Salesforce contact id starts with a lower case
+  //letter whereas all the other fields start with upper case so we need to set it explicitly
+  val salesforceIdName = "sfContactId__c"
```
davidfurey: Since `decapitalizeSfContactId` is private, why not make `salesforceIdName` private too?

common/src/main/scala/com/gu/zuora/encoding/CustomCodecs.scala
```diff
@@ -0,0 +1,57 @@
...
+  private def decapitalizeSfContactId(fn: String) =
```
davidfurey: Maybe just me, but I had to read it a few times to realise that 'fn' meant fieldName because I'm used to it meaning function.

common/src/main/scala/com/gu/zuora/model/SubscriptionData.scala
```diff
@@ -0,0 +1,31 @@
...
+
+case class SubscriptionData(ratePlanData: List[RatePlanData], subscription: Subscription)
+
+case class Subscription(
+  contractEffectiveDate: LocalDate,
+  contractAcceptanceDate: LocalDate,
+  termStartDate: LocalDate,
+  autoRenew: Boolean = true,
+  //scalastyle:off magic.number
+  initialTerm: Int = 12,
```
davidfurey: You can add the `// scalastyle:off magic.number` to the end of the line that you want it off for rather than turning it on and off.
`initialTerm: Int = 12, //scalastyle:off magic.number`
`renewalTerm: Int = 12, //scalastyle:off magic.number`

common/src/main/scala/com/gu/zuora/encoding/CapitalizationEncoder.scala
```diff
@@ -0,0 +1,34 @@
...
+import com.gu.helpers.StringExtensions._
+import com.gu.i18n.Currency
+import io.circe.generic.decoding.DerivedDecoder
+import io.circe.generic.encoding.DerivedObjectEncoder
+import io.circe.generic.semiauto._
+import io.circe.{Decoder, Encoder, JsonObject, ObjectEncoder}
+import shapeless.Lazy
+
+object CapitalizationEncoder {
+  implicit val encodeCurrency: Encoder[Currency] = Encoder.encodeString.contramap[Currency](_.iso)
```
rupertbates: It should be deleted actually!

common/src/main/scala/com/gu/zuora/encoding/CapitalizationEncoder.scala
rupertbates: About the same? I'm going to leave it as is I think

common/src/main/scala/com/gu/zuora/encoding/CustomCodecs.scala
rupertbates: 👍

common/src/main/scala/com/gu/zuora/encoding/CustomCodecs.scala
rupertbates: 👍

common/src/main/scala/com/gu/zuora/model/SubscriptionData.scala
rupertbates: Yes but this way you save one character! 

I think it might be good to switch the magic number inspection off, it's  a very blunt instrument currently

davidfurey:
👍

PR #11, opened 2017-05-11
Create a Scalatest tag for integration tests

rupertbates:
There are a number of integration tests in the project which talk to real services, these are very useful for real end to end testing, but slow to run and prone to failures if any of the services are playing up.

This PR adds a ScalaTest tag and annotation 'IntegrationTest' which can be applied to individual tests (the tag) or whole specs (the annotation) to stop them running when the sbt test command is invoked; it also tags the appropriate tests/specs with these tags.

### DIFF
--- a/monthly-contributions/src/test/scala/com/gu/support/workers/CreateZuoraSubscriptionSpec.scala
+++ b/monthly-contributions/src/test/scala/com/gu/support/workers/CreateZuoraSubscriptionSpec.scala
@@ -7,13 +7,13 @@ import com.gu.okhttp.RequestRunners
 import com.gu.support.workers.Conversions.{FromOutputStream, StringInputStreamConversions}
 import com.gu.support.workers.Fixtures._
 import com.gu.support.workers.lambdas.CreateZuoraSubscription
-import com.gu.support.workers.model.SendThankYouEmailState
 import com.gu.zuora.ZuoraService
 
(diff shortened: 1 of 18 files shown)


jacobwinch:
Would be good to add some information on running the different types of tests into the README too.

👍

monthly-contributions/src/test/scala/com/gu/support/workers/CreateZuoraSubscriptionSpec.scala
```diff
@@ -7,11 +7,13 @@ import com.gu.okhttp.RequestRunners
 import com.gu.support.workers.Conversions.{FromOutputStream, StringInputStreamConversions}
 import com.gu.support.workers.Fixtures._
 import com.gu.support.workers.lambdas.CreateZuoraSubscription
+import com.gu.test.tags.annotations.IntegrationTest
 import com.gu.zuora.ZuoraService
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 
+@IntegrationTest
```
jacobwinch: Nice that we can apply this to a full spec

rupertbates:
> Would be good to add some information on running the different types of tests into the README too.
> 
> 👍

@jacobwinch agreed, I've added something to the ReadMe

PR #12, opened 2017-05-12
Replace the scheduled Salesforce Auth token fetch

rupertbates:
Instead use the issued_at time of the token to check whether we should refresh it or not.

This simplifies the logic and also allows us to remove Akka from the project. 

To handle the concurrency & mutable state I'm using [ScalaSTM](https://nbronson.github.io/scala-stm/) which is really nice.

### DIFF
--- a/common/src/main/scala/com/gu/salesforce/SalesforceService.scala
+++ b/common/src/main/scala/com/gu/salesforce/SalesforceService.scala
@@ -1,12 +1,11 @@
 package com.gu.salesforce
 
-import akka.actor.{ActorSystem, Cancellable}
-import akka.agent.Agent
 import com.gu.config.Configuration
 import com.gu.helpers.{Retry, WebServiceHelper}
 import com.gu.okhttp.RequestRunners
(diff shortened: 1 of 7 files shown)


common/src/main/scala/com/gu/salesforce/SalesforceService.scala
```diff
@@ -48,32 +48,32 @@ class SalesforceService(config: SalesforceConfig, client: FutureHttpClient)(impl
...
-      sendAuth(authentication)
-      authentication
-    })
+  private val service = new AuthService(Configuration.salesforceConfig, RequestRunners.configurableFutureRunner(10.seconds))
+  private val authRef = Ref[Option[Authentication]](None)
+
+  def getAuth: Future[Authentication] = authRef.single() match {
+    case Some(authentication) =>
+      if (authentication.isStale)
+        fetchAuth //Fetch a new auth token but return the one we currently have
```
jacobwinch: 👍

jacobwinch:
Looks good to me

PR #16, opened 2017-05-19
Add Encryption

rupertbates:
Protect user data as it is passed through the various stages of the step functions by encrypting it using the [AWS KMS API](https://docs.aws.amazon.com/kms/latest/developerguide/programming-top.html).

This encryption can be switched off for debugging purposes by setting the aws.useEncryption config setting to false.

This PR relies on new config settings which will need to be fetched from S3


common/src/main/scala/com/gu/support/workers/encoding/Encryption.scala
```diff
@@ -0,0 +1,70 @@
...
+      //http://docs.aws.amazon.com/kms/latest/developerguide/workflow.html
+      kms.encrypt(req).getCiphertextBlob.array()
+    } catch {
+      case e: AWSKMSException => throw new EncryptionException("Error while encrypting data", e)
+    }
+  }
+
+  override def decrypt(data: Array[Byte]): String = {
+    val byteBuffer = ByteBuffer.wrap(data)
+    val req2 = new DecryptRequest().withCiphertextBlob(byteBuffer)
```
davidfurey: Probably rename `req2` to `req`?

PR #20, opened 2017-05-26
Error handling

rupertbates:
Documentation for error handling


docs/error-handling.md
```diff
@@ -0,0 +1,297 @@
...
+which are permanent failures. The rest of this document is an attempt to do that analysis for the Monthly contributions step function
+
+## Logging
+Most errors should be logged to Sentry so they can be investigated. However transient ones such as timeouts probably don't need to be logged in Sentry and can just be logged in Cloudwatch. This is also the case for Stripe card errors.
+
+## Types of errors
+
+### General errors
+
+<table>
```
svillafe: In the future, if you want to avoid html tabes, markdown also accepts them in [this format](http://www.tablesgenerator.com/markdown_tables).

common/src/main/scala/com.gu.okhttp/package.scala
```diff
@@ -9,7 +9,7 @@ import scala.concurrent.{Future, Promise}
 
 package object okhttp {
 
-  implicit class RickOkHttpClient(client: OkHttpClient) extends LazyLogging {
+  implicit class RichOkHttpClient(client: OkHttpClient) extends LazyLogging {
```
jacobwinch: 👍

common/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
```diff
@@ -0,0 +1,38 @@
...
+import com.amazonaws.services.kms.model._
+import com.gu.helpers.WebServiceHelperError
+import io.circe.ParsingFailure
+
+object RetryImplicits {
+
+  implicit class RetryConversions(val throwable: Throwable) {
+    def asRetryException: RetryException = throwable match {
+      //timeouts and 500s
+      case _: SocketTimeoutException |
```
jacobwinch: 👍

jacobwinch:
This looks great 👍 

My only slight concern is that the unlimited retries on SocketTimeoutExceptions could (theoretically) allow us to repeatedly create subscriptions (and re-charge customers) in the event of read timeouts from Zuora. Practically speaking this is mitigated considerably by the increased timeout for ZuoraService - but we might want to add a check for this to guard against it in a separate PR.

PR #22, opened 2017-05-31
Remove io.circe.generic.auto._

davidfurey:
The combination of some explicit encoder/decoders was becoming
confusing.  This commit adds lots of boilerplate but reduces
uncertainty.


common/src/main/scala/com/gu/support/workers/encoding/Helpers.scala
```diff
@@ -0,0 +1,15 @@
+package com.gu.support.workers.encoding
+
+import com.gu.zuora.encoding.CapitalizationEncoder._
+import io.circe.generic.decoding.DerivedDecoder
+import io.circe.generic.encoding.DerivedObjectEncoder
+import io.circe.generic.semiauto._
+import shapeless.Lazy
+
+object Helpers {
```
rupertbates: This is excellent! Good work 👍

rupertbates:
Looks a lot better now

PR #30, opened 2017-06-26
Connect function with mda

svillafe:
# Why are you doing this?

We want to connect the monthly contribution's state machine with the Members Data API. Therefore, in this PR we are introducing a new lambda function which handles the new step to make the PUT request to Members Data API. Since we didn't support PUT request before, we are adding that logic also. Finally, we are adding the configuration for this new step in the Configuration.scala file.

### DIFF
--- a/common/src/main/scala/com/gu/membersDataAPI/MembersDataService.scala
+++ b/common/src/main/scala/com/gu/membersDataAPI/MembersDataService.scala
@@ -22,7 +22,7 @@ object MembersDataServiceConfig {
 
 case class MembersDataServiceConfig(url: String, apiKey: String)
 
-case class ErrorResponse(error: String) extends Throwable
+case class ErrorResponse(message: String, details: String, statusCode: Int) extends Throwable
 
 object ErrorResponse {
   implicit val codec: Codec[ErrorResponse] = deriveCodec
--- a/common/src/main/scala/com/gu/membersDataAPI/MembersDataService.scala
+++ b/common/src/main/scala/com/gu/membersDataAPI/MembersDataService.scala
@@ -5,9 +5,18 @@ import com.gu.okhttp.RequestRunners
 import com.gu.okhttp.RequestRunners.FutureHttpClient
 import com.gu.support.workers.encoding.Codec
 import com.gu.support.workers.encoding.Helpers.deriveCodec
+import com.typesafe.config.Config
 
 import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration._
+
+object MembersDataServiceConfig {
+  def fromConfig(config: Config): MembersDataServiceConfig =
+    MembersDataServiceConfig(
+      url = config.getString("membersDataAPI.url"),
+      apiKey = config.getString("membersDataAPI.apiKey")
+    )
+}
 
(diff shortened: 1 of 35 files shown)


monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -14,7 +14,7 @@ import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
 class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
-  extends ServicesHandler[CreatePaymentMethodState, CreateSalesforceContactState](servicesProvider) with LazyLogging {
+    extends ServicesHandler[CreatePaymentMethodState, CreateSalesforceContactState](servicesProvider) with LazyLogging {
```
rupertbates: This formatting looks a bit weird, is this a Scalariform thing? Can we make it better?

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/UpdateMembersDataAPI.scala
```diff
@@ -0,0 +1,27 @@
...
+import com.gu.config.Configuration
+import com.gu.emailservices.EmailService
+import com.gu.membersDataAPI.MembersDataService
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import com.gu.support.workers.encoding.StateCodecs._
+
+import scala.concurrent.Future
+
+class UpdateMembersDataAPI(membersDataService: MembersDataService)
```
rupertbates: Are you supporting test users in the UAT environment? I think this Lambda should inherit from ServicesHandler and take its service dependency from there in the same way that CreateSalesforceContact does, this can then take care of passing in a service with the correct configuration (UAT, PROD or DEV). See ServiceProvider for where this is all set up for the other services

common/src/main/scala/com/gu/membersDataAPI/MembersDataService.scala
```diff
@@ -0,0 +1,43 @@
...
+      url = config.getString("membersDataAPI.url"),
+      apiKey = config.getString("membersDataAPI.apiKey")
+    )
+}
+
+case class MembersDataServiceConfig(url: String, apiKey: String)
+
+case class ErrorResponse(error: String) extends Throwable
+
+object ErrorResponse {
```
rupertbates: Is this a real response you are expecting from the api?

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
davidfurey: Apparently this conforms to the Scala style guide: http://docs.scala-lang.org/style/declarations.html#classes

rupertbates: Oh it's ok it just looks weird in the github diff, it's fine in the actual file.

PR #31, opened 2017-06-28
Fix membership data api update

davidfurey:
We discovered a few problems with the membership data api update, this fixes them.

screen shot 2017-06-28 at 16 50 14


svillafe:
After doing an end-to-end test in PROD it looks that is working. The step functions ran without failure.
picture 292

Answer fo
```JSON
{"userId":"17916359","recurringContributionPaymentPlan":"Monthly Contribution","contentAccess":{"member":false,"paidMember":false,"recurringContributor":true}}
```

PR #33, opened 2017-07-04
Adding CloudWatch's metrics

svillafe:
# Why are you doing this? 

We would like to have and analyse metrics for our signup process of different products. In this PR we are adding a basic schema of metrics and some metrics in the lambda functions.


common/src/main/scala/com/gu/monitoring/ErrorMetrics.scala
```diff
@@ -0,0 +1,19 @@
...
+
+  //Error's Dimensions
+  val stepName: String
+  val serviceName: String
+  val errorCode: String
+
+  val errorDimension: Seq[Dimension] =
+    Seq(new Dimension()
+        .withName("StepName").withValue(stepName)
+        .withName("ServiceName").withValue(serviceName)
```
davidfurey: Does this not overwrite line 14?

common/src/main/scala/com/gu/monitoring/ProductMetrics.scala
```diff
@@ -0,0 +1,19 @@
+package com.gu.monitoring
+
+import com.amazonaws.services.cloudwatch.model.Dimension
+
+trait ProductMetrics extends CloudWatch {
+
+  //Product's Dimensions
+  val productName: String
```
davidfurey: Any reason for these to be `val`s but `mandatoryDimensions` is a `def`?

common/src/main/scala/com/gu/monitoring/products/RecurringContributionsMetrics.scala
```diff
@@ -0,0 +1,20 @@
+package com.gu.monitoring.products
+
+import com.gu.config.Configuration
+import com.gu.monitoring.ProductMetrics
+
+class RecurringContributionsMetrics(val paymentMethod: String,
+                                    val subscriptionPeriod: String) extends ProductMetrics {
+
+  override val productName = "RecurringContributor"
+  override val stage: String = Configuration.stage
```
davidfurey: Type specified here but not on line 9?

common/src/main/scala/com/gu/monitoring/products/RecurringContributionsMetrics.scala
```diff
@@ -0,0 +1,20 @@
+package com.gu.monitoring.products
+
+import com.gu.config.Configuration
+import com.gu.monitoring.ProductMetrics
+
+class RecurringContributionsMetrics(val paymentMethod: String,
```
davidfurey: I think you could put `override` for these two `val`s so that it is clear that they are used in the parent class.  Otherwise they look unused

common/src/main/scala/com/gu/monitoring/ErrorMetrics.scala
```diff
@@ -0,0 +1,19 @@
+package com.gu.monitoring
+
+import com.amazonaws.services.cloudwatch.model.Dimension
+
+trait ErrorMetrics extends CloudWatch {
```
davidfurey: Is this trait used?

common/src/main/scala/com/gu/monitoring/CloudWatch.scala
```diff
@@ -0,0 +1,62 @@
...
+import com.gu.config.Configuration
+import com.gu.monitoring.CloudWatch.cloudwatch
+import com.typesafe.scalalogging.LazyLogging
+
+class CloudWatch(metrics: Seq[Dimension]) extends LazyLogging {
+
+  val application = "SupportWorkers"
+
+  //Common Dimensions
+  val commonDimensions: Seq[Dimension] =
```
davidfurey: Since there is now just one common dimension, maybe this could become
`val stageDimension: Dimension = ...`

common/src/main/scala/com/gu/monitoring/CloudWatch.scala
```diff
@@ -0,0 +1,62 @@
...
+
+object CloudWatch {
+
+  lazy val cloudwatch: AmazonCloudWatchAsync = AmazonCloudWatchAsyncClient.asyncBuilder
+    .withCredentials(CredentialsProvider)
+    .withRegion(EU_WEST_1).build()
+}
+
+object CloudWatchHealth {
+  var hasPushedMetricSuccessfully = false
```
davidfurey: This value is never read, so what is it for?

common/src/main/scala/com/gu/monitoring/MemberDataAPIMetrics.scala
```diff
@@ -0,0 +1,13 @@
+package com.gu.monitoring
+
+import Metrics._
+class MembersDataAPIMetrics(val name: String) extends CloudWatch(Seq(
+  productName(name)
```
davidfurey: I don't think the extra newlines make this easier to read than
`class MembersDataAPIMetrics(val name: String) extends CloudWatch(Seq(productName(name))) {`

common/src/main/scala/com/gu/monitoring/CloudWatch.scala
```diff
@@ -0,0 +1,62 @@
...
+import com.amazonaws.handlers.AsyncHandler
+import com.amazonaws.regions.Regions.EU_WEST_1
+import com.amazonaws.services.cloudwatch.{AmazonCloudWatchAsync, AmazonCloudWatchAsyncClient, AmazonCloudWatchAsyncClientBuilder}
+import com.amazonaws.services.cloudwatch.model.{Dimension, MetricDatum, PutMetricDataRequest, PutMetricDataResult}
+import com.gu.aws.CredentialsProvider
+import com.gu.config.Configuration
+import com.gu.monitoring.CloudWatch.cloudwatch
+import com.typesafe.scalalogging.LazyLogging
+
+class CloudWatch(metrics: Seq[Dimension]) extends LazyLogging {
```
davidfurey: Could change this to `class CloudWatch(metrics: Dimension*)` to avoid the `Seq` where you extend this class

common/src/main/scala/com/gu/monitoring/CloudWatch.scala
```diff
@@ -0,0 +1,62 @@
...
+  val commonDimensions: Seq[Dimension] =
+    Seq(new Dimension().withName("Stage").withValue(Configuration.stage))
+
+  val allDimensions: Seq[Dimension] = commonDimensions ++ metrics
+
+  trait LoggingAsyncHandler extends AsyncHandler[PutMetricDataRequest, PutMetricDataResult] {
+    def onError(exception: Exception) {
+      logger.info(s"CloudWatch PutMetricDataRequest error: ${exception.getMessage}}")
+    }
+    def onSuccess(request: PutMetricDataRequest, result: PutMetricDataResult) {
```
davidfurey: If this and `onError` are overriding or implementing a method in AsyncHandler use the `override` keyword to make this obvious.

common/src/main/scala/com/gu/monitoring/MemberDataAPIMetrics.scala
```diff
@@ -0,0 +1,13 @@
...
+)) {
+
+  def putMembersDataAPIUpdated(): Unit = put(s"$name-members-data-api-updated")
```
davidfurey: This shouldn't return `Unit`, it should return a `Future` which completes when the `put` request has completed.

common/src/main/scala/com/gu/monitoring/products/RecurringContributionsMetrics.scala
```diff
@@ -0,0 +1,21 @@
...
+class RecurringContributionsMetrics(
+  val method: String,
+  val period: String
+) extends CloudWatch(Seq(
+  productName("RecurringContributor"),
+  paymentMethod(method),
+  subscriptionPeriod(period)
+)) {
+
+  def putContributionSignUpStartProcess(): Unit = put(s"monthly-contributor-sign-up-start")
```
davidfurey: This (and  `putZuoraAccountCreated `) shouldn't return Unit, it should return a Future which completes when the put request has completed.

common/src/main/scala/com/gu/monitoring/products/RecurringContributionsMetrics.scala
```diff
@@ -0,0 +1,21 @@
+package com.gu.monitoring.products
+
+import com.gu.monitoring._
+import Metrics._
+class RecurringContributionsMetrics(
+  val method: String,
```
davidfurey: Do these need the `val` keyword?

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -20,10 +22,18 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
 
   override protected def servicesHandler(state: CreatePaymentMethodState, context: Context, services: Services) = {
     logger.debug(s"CreatePaymentMethod state: $state")
+
     val paymentMethod = state.paymentFields match {
-      case Left(stripe) => createStripePaymentMethod(stripe, services.stripeService)
-      case Right(payPal) => createPayPalPaymentMethod(payPal, services.payPalService)
+      case Left(stripe) => {
```
davidfurey: You don't need the `{` here

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -19,6 +20,10 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
 
   override protected def servicesHandler(state: CreateZuoraSubscriptionState, context: Context, services: Services) =
     services.zuoraService.subscribe(buildSubscribeRequest(state)).map { response =>
+      state.paymentMethod.`type` match {
```
davidfurey: No need for a pattern match here, `if`/`else` would be clearer for something this simple.

Also, why not just ```putCloudWatchMetrics(state.paymentMethod.`type`)```?

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/UpdateMembersDataAPI.scala
```diff
@@ -14,6 +17,10 @@ class UpdateMembersDataAPI(servicesProvider: ServiceProvider = ServiceProvider)
   def this() = this(ServiceProvider)
 
   override protected def servicesHandler(state: UpdateMembersDataAPIState, context: Context, services: Services): Future[Unit] = {
-    services.membersDataService.update(state.user.id, state.user.isTestUser).map(_ => Unit)
+    services.membersDataService.update(state.user.id, state.user.isTestUser).map(_ => putCloudWatchMetrics)
```
davidfurey: The `Future` that this returns will complete before `putCloudWatchMetrics` is complete, so the lambda may terminate before the cloudwatch request has happened.  If you modify putCloudWatchMetrics to return a `Future` you can use `flatMap` instead of `map` which will fix this problem

common/src/main/scala/com/gu/monitoring/ProductMetrics.scala
```diff
@@ -0,0 +1,9 @@
+package com.gu.monitoring
+
+import com.amazonaws.services.cloudwatch.model.Dimension
+
+object Metrics {
```
davidfurey: I suppose this is really `Dimensions`?

common/src/main/scala/com/gu/monitoring/CloudWatch.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  val application = "SupportWorkers"
+
+  //Common Dimensions
+  val stage: String
+  val backend: String
+
+  lazy val commonDimensions: Dimension = new Dimension()
+    .withName("Stage").withValue(stage)
+    .withName(backend).withValue(backend)
```
svillafe: https://github.com/guardian/support-workers/pull/33/commits/853e8cc5e9ffd1d7a4abd2b7b12a6783812aeb81#diff-276d3eae1d78d5642859c7e0e5f61009R8

common/src/main/scala/com/gu/monitoring/CloudWatch.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  def mandatoryDimensions:Seq[Dimension]
```
svillafe: Renamed it in: f92b211126145ab28f01ded6c633a086d5bf7783

svillafe:
Working in DEV via tests:
picture 298
picture 299

PR #34, opened 2017-07-04
Json wrapper

rupertbates:
AWS Step Functions expect to be passed valid Json, as we want to encrypt the whole of the
state, we need to Base64 encode it and then wrap it in a Json 'wrapper' object.

This PR does that and then fixes all the tests

### DIFF
--- a/common/src/main/scala/com/gu/support/workers/encoding/Wrapper.scala
+++ b/common/src/main/scala/com/gu/support/workers/encoding/Wrapper.scala
@@ -30,7 +30,9 @@ object Wrapper {
   }
 
   def wrap[T](value: T)(implicit encoder: Encoder[T]): JsonWrapper =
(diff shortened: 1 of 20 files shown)


common/src/main/scala/com/gu/support/workers/encoding/Wrapper.scala
```diff
@@ -0,0 +1,38 @@
...
+import scala.io.Source
+import scala.util.Try
+
+/**
+ * AWS Step Functions expect to be passed valid Json, as we want to encrypt the whole of the
+ * state, we need to wrap it in a Json 'wrapper' object as a Base64 encoded String.
+ * This class helps with that
+ */
+object Wrapper {
+  implicit val jsonEncoder = deriveEncoder[JsonWrapper]
```
davidfurey: You could use Helpers.deriveCodec here

PR #38, opened 2017-07-11
Check for an existing contribution subscription before creating a new one

rupertbates:
## Why are you doing this?

The subscribe call to Zuora is not idempotent and it is possible for the request to time out (from the point of view of the step functions) whilst actually still succeeding in Zuora. This would result in us retrying and potentially creating multiple live subscriptions.

To get around this problem I have introduced a check to the CreateZuoraSubscription lambda so that if an existing monthly recurring subscription exists for the current user then we skip the account creation step and just move on to the next step in the monthly contribution flow.

[**Trello Card**](https://trello.com/c/1i1MhfIh/615-q2-test-2-check-for-existing-monthly-contribution-before-attempting-zuora-subscribe-call)

### DIFF
--- a/common/src/main/scala/com/gu/zuora/ZuoraService.scala
+++ b/common/src/main/scala/com/gu/zuora/ZuoraService.scala
@@ -4,6 +4,7 @@ import cats.syntax.either._
 import com.gu.helpers.WebServiceHelper
 import com.gu.okhttp.RequestRunners.FutureHttpClient
 import com.gu.zuora.model._
+import com.gu.zuora.model.response._
 import io.circe
 import io.circe.Decoder
 import io.circe.parser.decode
@@ -25,6 +26,14 @@ class ZuoraService(config: ZuoraConfig, client: FutureHttpClient, baseUrl: Optio
 
   def getAccount(accountNumber: String): Future[GetAccountResponse] =
     get[GetAccountResponse](s"accounts/$accountNumber")
+
+  def getAccountIds(identityId: String): Future[QueryResponse] = {
+    val queryData = QueryData(s"select AccountNumber from account where IdentityId__c = '$identityId'")
+    post[QueryResponse](s"action/query", queryData.asJson)
+  }
+
+  def getSubscriptions(accountId: String): Future[SubscriptionsResponse] =
+    get[SubscriptionsResponse](s"subscriptions/accounts/$accountId")
 
   def subscribe(subscribeRequest: SubscribeRequest): Future[List[SubscribeResponseAccount]] =
     post[List[SubscribeResponseAccount]](s"action/subscribe", subscribeRequest.asJson)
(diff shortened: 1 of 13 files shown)


common/src/test/scala/com/gu/zuora/ZuoraSpec.scala
```diff
@@ -12,17 +13,45 @@ import scala.concurrent.duration._
...
 class ZuoraSpec extends AsyncFlatSpec with Matchers with LazyLogging {
 
+  def uatService: ZuoraService = new ZuoraService(zuoraConfigProvider.get(true), RequestRunners.configurableFutureRunner(30.seconds))
+
   "ZuoraService" should "retrieve an account" in {
-    val zuoraService = new ZuoraService(Configuration.zuoraConfigProvider.get(), RequestRunners.configurableFutureRunner(10.seconds))
-    zuoraService.getAccount(Fixtures.accountNumber).map {
+    uatService.getAccount(Fixtures.accountNumber).map {
       response =>
+        logger.info(s"$response")
```
JustinPinner: I guess this (and below) `logger.info` statements can go from tests? Maybe in the next update?

JustinPinner:
This makes sense I think. Might want to get a second opinion, but otherwise 👍

common/src/main/scala/com/gu/zuora/ZuoraService.scala
```diff
@@ -26,12 +28,26 @@ class ZuoraService(config: ZuoraConfig, client: FutureHttpClient, baseUrl: Optio
...
+  def getAccountIds(identityId: String): Future[List[String]] = {
+    val queryData = QueryData(s"select AccountNumber from account where IdentityId__c = '$identityId'")
+    post[QueryResponse](s"action/query", queryData.asJson).map(_.records.map(_.AccountNumber))
+  }
+
+  def getSubscriptions(accountId: String): Future[List[Subscription]] =
+    get[SubscriptionsResponse](s"subscriptions/accounts/$accountId").map(_.subscriptions)
+
   def subscribe(subscribeRequest: SubscribeRequest): Future[List[SubscribeResponseAccount]] =
     post[List[SubscribeResponseAccount]](s"action/subscribe", subscribeRequest.asJson)
```
davidfurey: Not part of this PR, but there's no need for the string interpolation here (`s""`)

common/src/main/scala/com/gu/zuora/ZuoraService.scala
```diff
@@ -26,12 +28,26 @@ class ZuoraService(config: ZuoraConfig, client: FutureHttpClient, baseUrl: Optio
   def getAccount(accountNumber: String): Future[GetAccountResponse] =
     get[GetAccountResponse](s"accounts/$accountNumber")
 
+  def getAccountIds(identityId: String): Future[List[String]] = {
+    val queryData = QueryData(s"select AccountNumber from account where IdentityId__c = '$identityId'")
```
rupertbates: This isn't SQL it's ZOQL which is quite a bit more constrained, but I'll parse it to a Long first to be safe.

PR #40, opened 2017-07-18
Use the support-config project

rupertbates:
## Why are you doing this?
I have moved the classes related to Touchpoint config to a separate project [support-config](https://github.com/guardian/support-config) to ensure that we have consistency of concepts between this project and support-frontend.

This PR updates support-workers to use this new project

### NB: Need to check aws conf files on merging


common/src/main/scala/com/gu/config/Configuration.scala
```diff
@@ -19,41 +17,18 @@ object Configuration extends LazyLogging {
     .getOrElse("TRUE").toBoolean)
     .getOrElse(true) //Should we load config from S3
 
-  val stage = Stage(Option(System.getenv("GU_SUPPORT_WORKERS_STAGE")).getOrElse(Stages.DEV))
+  val stage = Stage.fromString(System.getenv("GU_SUPPORT_WORKERS_STAGE")).getOrElse(Stages.DEV)
```
davidfurey: I think `System.getenv("GU_SUPPORT_WORKERS_STAGE")` still needs to be wrapped in an Option since it can return `null`

rupertbates: It's fine because if the environment variable is null then the match fails in .fromString and we use the default.

PR #45, opened 2017-07-28
Add complete state

davidfurey:
## Why are you doing this?

Adding an extra step so that support-frontend can notify the user when the transaction has succeeded.  This is as soon as the Zuora step completes, since at this point the user will be billed.


rupertbates:
👍

PR #48, opened 2017-08-02
Improved error handling

davidfurey:
## Why are you doing this?

To provide immediate feedback to the customer so they know to try again now or later.

Also improves exception logging generally

### DIFF
--- a/common/src/main/scala/com/gu/zuora/model/response/Responses.scala
+++ b/common/src/main/scala/com/gu/zuora/model/response/Responses.scala
@@ -6,7 +6,6 @@ import com.gu.support.workers.exceptions.{RetryException, RetryNone}
 import io.circe.syntax._
 import io.circe.parser._
 import cats.implicits._
-import com.typesafe.scalalogging.LazyLogging
 
 sealed trait ZuoraResponse {
   def success: Boolean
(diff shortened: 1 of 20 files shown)


common/src/main/scala/com/gu/zuora/model/response/Responses.scala
```diff
@@ -14,12 +18,24 @@ object ZuoraError {
 
 case class ZuoraError(Code: String, Message: String)
 
-object ZuoraErrorResponse {
+object ZuoraErrorResponse extends LazyLogging {
```
rupertbates: Don't need to extend LazyLogging here I don't think

common/src/main/scala/com/gu/zuora/model/response/Responses.scala
```diff
@@ -14,12 +18,24 @@ object ZuoraError {
...
+    } else {
+      None
+    }
+  }
 }
 
-case class ZuoraErrorResponse(success: Boolean, errors: List[ZuoraError]) extends Throwable with ZuoraResponse {
-  def asRetryException: RetryException = new RetryNone(cause = this)
+case class ZuoraErrorResponse(success: Boolean, errors: List[ZuoraError])
+    extends Throwable(errors.asJson.noSpaces) with ZuoraResponse {
```
rupertbates: spaces2 might be easier to read when debugging than noSpaces

common/src/main/scala/com/gu/zuora/model/response/Responses.scala
davidfurey: Thanks, I forgot to remove that after some debugging.

common/src/main/scala/com/gu/zuora/model/response/Responses.scala
davidfurey: Good idea

PR #51, opened 2017-08-09
Refactor cfn templates

rupertbates:
## Why are you doing this?
The way we currently build our Cloud formation templates is a getting a bit unwieldy now that we have a lot of lambdas and some nested templates. 

Using a proper templating language (Mustache) makes it much easier to understand what is going on and reduces the amount of boilerplate. I have switched from Python to Node for the execution environment as it has better support for working with Mustache templates.

This PR just replaces the old system with the new one. After merging this I will move on to adding the new templates needed for the Digital Bundle.

There will also need to be some additional changes to get the new system to build in TeamCity, which I will probably add in to this PR.

[**Trello Card**](https://trello.com/c/dWpYddPv/792-new-cloudformation-templates-for-support-workers-digital-bundle)


JustinPinner:
Took me a few re-reads and much frowning to parse this into wetware, but I think I'm beginning to understand. 👍 from me (but please seek others, too!)

cloud-formation/src/templates/lambda.yaml
```diff
@@ -0,0 +1,15 @@
+  {{name}}:
+    Type: "AWS::Lambda::Function"
+    Description: "{{description}}"
+    Properties:
+      {{> environmentVariables}}
+
+      FunctionName: !Sub ${Stack}-ContributionCompletedLambda-${Stage}
```
davidfurey: Should the references to "CreatePaymentMethodLambda" not be `{{name}}` or similar?

cloud-formation/src/cfn.json
```diff
@@ -0,0 +1,32 @@
+{
```
davidfurey: Could this file be yaml?

cloud-formation/src/templates/lambda.yaml
rupertbates: 👍 Good spot

cloud-formation/src/templates/lambda.yaml
```diff
@@ -0,0 +1,15 @@
+  {{name}}:
+    Type: "AWS::Lambda::Function"
+    Description: "{{description}}"
+    Properties:
+      {{> environmentVariables}}
+
+      FunctionName: !Sub ${Stack}-{{name}}-${Stage}
+      Handler: "com.gu.support.workers.lambdas.ContributionCompleted::handleRequest"
```
davidfurey: I think `ContributionCompleted` shouldn't be hard-coded here

rupertbates: It's not any more, not sure why you are seeing it like that.

rupertbates: Ok fixed now

PR #52, opened 2017-08-16
Add logging of successful and failed contributions

davidfurey:
## Why are you doing this?

So we can create a cloudwatch dashboard of contributions


JustinPinner:
Are these logs intending to give us enough information to find and fix the error, or just to highlight problems more generally? I'm guessing the latter due to the absence of individual transaction identifiers, in which case 👍

PR #54, opened 2017-08-29
Fix for American Express

jacobwinch:
## Why are you doing this?

The monthly contributions flow currently fails if the user attempts to pay with American Express, because Zuora requires the card type to be sent as 'AmericanExpress', not 'American Express'.

See also:
https://github.com/guardian/membership-common/pull/371

[**Trello Card**](https://trello.com/c/lrELOTgh/810-fix-american-express)

## Changes

* Remove whitespace when creating CreditCardReferenceTransaction for Zuora

cc @davidfurey @rupertbates

### DIFF
--- a/monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
+++ b/monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
@@ -50,8 +50,9 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
       .createCustomer(stripe.userId, stripe.stripeToken)
       .map { stripeCustomer =>
         val card = stripeCustomer.card
+        val zuoraCardType = card.`type`.replaceAll(" ", "") // Zuora requires 'AmericanExpress' not 'American Express'
         CreditCardReferenceTransaction(card.id, stripeCustomer.id, card.last4,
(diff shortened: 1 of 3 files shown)


monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -50,8 +50,9 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
       .createCustomer(stripe.userId, stripe.stripeToken)
       .map { stripeCustomer =>
         val card = stripeCustomer.card
+        val zuoraCardType = card.`type`.replaceAll(" ", "") // Zuora requires 'AmericanExpress' not 'American Express'
```
rupertbates: Could maybe move this into the Stripe.Card class as there is another derived value in there already. I don't feel super strongly about it though if you'd rather keep it here.

rupertbates: I'm assuming you've checked that there are no types for which we'd want to retain spaces?

jacobwinch: Good idea - I've also linked to the docs which lists the types that Zuora will accept

jacobwinch:
I've deployed this to CODE and confirmed that the Step Function completes successfully with both Amex and Visa so I'm going to ship this...

PR #56, opened 2017-09-01
Annual contributions

rupertbates:
## Why are you doing this?
To enable us to bill supporters annually as well as monthly.

### NB
* This change depends on [this PR](https://github.com/guardian/support-models/pull/13) in support-models. It is backwardly compatible with the existing version of support-frontend, because any request received which does not specify the billing period will default to Monthly.
* This change requires updates to support-worker private conf 

[**Trello Card**](https://trello.com/c/qF6I86xx/878-uk-annual-contributions)

### DIFF
--- a/common/src/main/scala/com/gu/zuora/ZuoraConfig.scala
+++ b/common/src/main/scala/com/gu/zuora/ZuoraConfig.scala
@@ -1,16 +1,36 @@
 package com.gu.zuora
 
 import com.gu.support.config.{Stage, TouchpointConfig, TouchpointConfigProvider}
+import com.gu.support.workers.model.{Annual, BillingPeriod}
 import com.typesafe.config.Config
 
-case class ZuoraConfig(url: String, username: String, password: String, productRatePlanId: String, productRatePlanChargeId: String) extends TouchpointConfig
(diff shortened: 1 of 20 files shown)


common/src/main/scala/com/gu/zuora/ZuoraConfig.scala
```diff
@@ -1,16 +1,36 @@
...
+case class ZuoraConfig(
+  url: String,
+  username: String,
+  password: String,
+  monthlyContribution: ZuoraContributionConfig,
+  annualContribution: ZuoraContributionConfig)
+  extends TouchpointConfig {
+  def configForBillingPeriod(billingPeriod: BillingPeriod): ZuoraContributionConfig =
+    billingPeriod match {
+      case _: Annual.type => annualContribution
```
davidfurey: I think this can be `case Annual => annualContribution`

project/Dependencies.scala
```diff
@@ -1,5 +1,6 @@
 import sbt._
 
+//noinspection TypeAnnotation
```
davidfurey: What does this do?

rupertbates: It disables the 'Type annotation required' inspection for the class. I really don't think it's necessary on the Dependencies.

PR #59, opened 2017-09-25
Pass the correct parameters into the Salesforce upsert

rupertbates:
## Why are you doing this?
To ensure that we are passing the same fields to Salesforce as we are to Zuora, this reduces the chance of data getting lost in syncs between the two systems.

[**Trello Card**](https://trello.com/c/7XRbUJkC/916-add-additional-fields-into-the-data-we-send-to-salesforce)

### DIFF
--- a/common/src/test/scala/com/gu/salesforce/Fixtures.scala
+++ b/common/src/test/scala/com/gu/salesforce/Fixtures.scala
@@ -5,17 +5,19 @@ object Fixtures {
   val salesforceId = "003g000001UnFItAAN"
   val email = "6cwdm8aler7z9r6nwbc@gu.com"
   val name = "6cWdM8AlER7z9R6nWBc"
+  val dummyValue = "Dummy Value"
   val allowMail = false
   val upsertJson =
     s"""{
       "newContact": {
         "IdentityID__c": "$idId",
-         "Email": "$email",
-         "FirstName": "$name",
-         "LastName": "$name",
-         "Allow_Membership_Mail__c": $allowMail,
-         "Allow_3rd_Party_Mail__c": $allowMail,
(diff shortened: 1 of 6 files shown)


JustinPinner:
This looks sensible. One question - we only seem to be testing with a `None` in the `State` field. As this looks like the first time we've used an `Option` here, is it worth testing with a `Some` value too?

common/src/test/scala/com/gu/salesforce/Fixtures.scala
```diff
@@ -5,17 +5,19 @@ object Fixtures {
...
-         "Email": "$email",
-         "FirstName": "$name",
-         "LastName": "$name",
-         "Allow_Membership_Mail__c": $allowMail,
-         "Allow_3rd_Party_Mail__c": $allowMail,
-         "Allow_Guardian_Related_Mail__c": $allowMail
+        "Email": "$email",
+        "FirstName": "$name",
+        "LastName": "$name",
+        "MailingCountry": "$dummyValue",
```
Ap0c: Why does this include country but not state?

rupertbates: Because if the state is missing as it will be in the uk we would send nothing and this is what this tests. As @JustinPinner suggested I'll add another test for the case where we do have a state.

JustinPinner:
👍

pvighi:
The only additional comment I have is should these be mailing details or billing details in salesforce?

rupertbates:
@pvighi I've put them into Mailing details because that's what membership-frontend does currently
https://github.com/guardian/membership-frontend/blob/master/frontend/app/services/SalesforceService.scala#L60

PR #60, opened 2017-09-29
submit acquisition events to Ophan

Mullefa:
## Why are you doing this?

To submit acquisition events to Ophan.

## Changes

* add a new step function to be executed after the contribution has been made. It builds the acquisition event from the input state, and submits it to Ophan

See [this](https://github.com/guardian/support-models/pull/14) pull request in `supporter-models` for the respective updates to the models.


common/src/main/scala/com/gu/akka/implicits/package.scala
```diff
@@ -0,0 +1,11 @@
+package com.gu.akka
```
rupertbates: I'm not massively keen on bringing Akka into the project, is it just handling http? Could we use OkHttp and the [WebServiceHelper](https://github.com/guardian/support-workers/blob/master/common/src/main/scala/com/gu/helpers/WebServiceHelper.scala) class in the same way as [ZuoraService](https://github.com/guardian/support-workers/blob/master/common/src/main/scala/com/gu/zuora/ZuoraService.scala) or [SalesforceService](https://github.com/guardian/support-workers/blob/master/common/src/main/scala/com/gu/salesforce/SalesforceService.scala) do?

desbo:
I've updated the CloudFormation templates to include the event submission - executed as one of the parallel tasks in the step function setup.

rupertbates:
Looks good, have you tested on CODE?

desbo:
@rupertbates now successfully tested on CODE: 

screen shot 2017-10-09 at 16 43 39

PR #62, opened 2017-10-04
Take encryption settings from the request state, not config

rupertbates:
## Why are you doing this?
To make it easier to switch encryption on and off without a risk of support-frontend and support-workers getting out of sync, this PR moves the setting which determines whether a request is encrypted or not into the request state rather than the config setting.

This means that support-frontend is now the only place where we need to update the encryption setting and support-workers will handle whatever it is sent. This would also allow us to send a mix of encrypted and unencrypted requests if for instance we wanted to keep test user requests unencrypted

Depends on PR https://github.com/guardian/support-models/pull/17

[**Trello Card**](https://trello.com/c/PxnJi97v/948-switch-on-encryption-in-the-support-stack)

### DIFF
--- a/common/src/main/scala/com/gu/zuora/encoding/CustomCodecs.scala
+++ b/common/src/main/scala/com/gu/zuora/encoding/CustomCodecs.scala
@@ -71,7 +71,7 @@ trait ModelsCodecs {
   implicit val executionErrorCodec: Codec[ExecutionError] = deriveCodec
   implicit val jsonWrapperDecoder: Decoder[JsonWrapper] = Decoder
     .forProduct3("state", "error","encrypted")(JsonWrapper.apply)
-    .or(Decoder.forProduct2("state", "error")((s: String, e: Option[ExecutionError]) => JsonWrapper(s, e)))
+    .or(Decoder.forProduct2("state", "error")((s: String, e: Option[ExecutionError]) => JsonWrapper(s, e, encrypted = false)))
   implicit val jsonWrapperEncoder: Encoder[JsonWrapper] = deriveEncoder
 
 }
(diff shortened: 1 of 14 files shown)


common/src/main/scala/com/gu/zuora/encoding/CustomCodecs.scala
```diff
@@ -68,6 +67,13 @@ trait ModelsCodecs {
     .forProduct3("amount", "currency", "billingPeriod")(Contribution.apply)
     .or(Decoder.forProduct2("amount", "currency")((a: BigDecimal, c: Currency) => Contribution(a, c)))
   implicit val encodeContribution: Encoder[Contribution] = deriveEncoder
+
+  implicit val executionErrorCodec: Codec[ExecutionError] = deriveCodec
+  implicit val jsonWrapperDecoder: Decoder[JsonWrapper] = Decoder
+    .forProduct3("state", "error","encrypted")(JsonWrapper.apply)
```
davidfurey: If encrypted was Option[Boolean] I don't think you'd need to complicate the decoder with the forProduct

rupertbates: That's true, but it's not really an optional field, this is just a consequence of having to change 2 systems. I guess you could argue it either way but I thought it was better to hide the complexity in the decoder and keep it simple everywhere else.

davidfurey: Fair enough.  Another option would be to remove the default value from support-models, and put it here.  Once this change has been rolled out we can follow up with a PR to remove the forPoduct2 making it a required field.

rupertbates: Yes I think that probably would be better, it did occur to me that defining the default in support-models might not be the best place.

rupertbates: Done

PR #68, opened 2017-10-20
adds cloudwatch metrics for each lambda function

michaelwmcnamara:
There are currently a small number of cloudwatch metrics sent by this application under the "support-worker" namespace. This change adds some more.
The reason for these new metrics is so that I create some composite metrics in a tool called "librato" that will calculate the ratio of successful completions of monthly payments to number of payments attempted and trigger an alert if this falls below a particular rate.

Why here when there are lambda metrics and state-machine metrics in AWS already?
1) It is not currently possible to import metrics under the "state" namespace into Librato. They are just not recognised.
2) Should this change, there is still a problem with the metrics in the "state" namespace. Their metric name includes a versioning id. This changes each time a new version is deployed, so I'd have to be continually updating the "Lambda" namespace are divded by type of metric. The actual function they represent is set as a tag. Unfortunately, for some reason not all of these tags are being pulled through for the librato metrics (I suspect we have exceeded the maximum number of tags librato can handle for a single metric with these metrics).

This change is part of the following [trello card](https://trello.com/c/tmz6kxm2/919-make-sure-client-side-logging-and-metrics-up-to-scratch).

### DIFF
--- a/monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
+++ b/monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
@@ -22,10 +22,10 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
 
   override protected def servicesHandler(state: CreatePaymentMethodState, context: Context, services: Services) = {
     logger.debug(s"CreatePaymentMethod state: $state")
-    putStartup(state.paymentFields)
+    putMetric(state.paymentFields, startupMetric)
     for {
       paymentMethod <- createPaymentMethod(state.paymentFields, services)
-      _ <- putPaymentMethodCreated(state.paymentFields)
+      _ <- putMetric(state.paymentFields, paymentMethodCreatedMetric)
     } yield getCreateSalesforceContactState(state, paymentMethod)
   }
 
@@ -47,7 +47,7 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
       state.acquisitionData
     )
 
-  def createStripePaymentMethod(stripe: StripePaymentFields, stripeService: StripeService): Future[CreditCardReferenceTransaction] = {
+  def createStripePaymentMethod(stripe: StripePaymentFields, stripeService: StripeService): Future[CreditCardReferenceTransaction] =
     stripeService
       .createCustomer(stripe.userId, stripe.stripeToken)
       .map { stripeCustomer =>
@@ -55,25 +55,17 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
         CreditCardReferenceTransaction(card.id, stripeCustomer.id, card.last4,
           CountryGroup.countryByCode(card.country), card.exp_month, card.exp_year, card.zuoraCardType)
       }
-  }
 
-  def createPayPalPaymentMethod(payPal: PayPalPaymentFields, payPalService: PayPalService): Future[PayPalReferenceTransaction] = {
+  def createPayPalPaymentMethod(payPal: PayPalPaymentFields, payPalService: PayPalService): Future[PayPalReferenceTransaction] =
     payPalService
       .retrieveEmail(payPal.baid)
       .map(PayPalReferenceTransaction(payPal.baid, _))
-  }
 
-  private def putStartup(paymentType: Either[StripePaymentFields, PayPalPaymentFields]) =
+  private def putMetric(paymentType: Either[StripePaymentFields, PayPalPaymentFields], f: String => Future[Unit]) =
     if (paymentType.isLeft)
-      startupMetric("stripe")
+      f("stripe")
     else
-      startupMetric("paypal")
-
-  private def putPaymentMethodCreated(paymentType: Either[StripePaymentFields, PayPalPaymentFields]) =
-    if (paymentType.isLeft)
-      paymentMethodCreatedMetric("stripe")
-    else
-      paymentMethodCreatedMetric("paypal")
+      f("paypal")
 
   def startupMetric(paymentMethod: String): Future[Unit] =
     new RecurringContributionsMetrics(paymentMethod, "monthly")
--- a/monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
+++ b/monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
@@ -25,7 +25,7 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
 
     for {
       paymentMethod <- createPaymentMethod(state.paymentFields, services)
-      _ <- putMetric(state.paymentFields)
+      _ <- putStartupMetric(state.paymentFields)
     } yield getCreateSalesforceContactState(state, paymentMethod)
   }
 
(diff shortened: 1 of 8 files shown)


monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -53,21 +47,35 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
...
+      putCloudWatchStartupMetric("paypal")
+
+  def putCloudWatchStartupMetric(paymentMethod: String): Future[Unit] =
+    new RecurringContributionsMetrics(paymentMethod, "monthly")
+      .putContributionSignUpStartProcess().recover({ case _ => () })
 
-  def putCloudWatchMetrics(paymentMethod: String): Future[Unit] =
+  def putCloudWatchCreatePaymentMethodMetric(paymentMethod: String): Future[Unit] =
     new RecurringContributionsMetrics(paymentMethod, "monthly")
       .putContributionSignUpStartProcess().recover({ case _ => () })
```
jacobwinch: Should this be: putCreatePaymentMethod()?

michaelwmcnamara: That does sound better. Good idea!

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreateSalesforceContact.scala
```diff
@@ -44,4 +46,15 @@ class CreateSalesforceContact extends ServicesHandler[CreateSalesforceContactSta
       response.ContactRecord,
       state.acquisitionData
     )
+
+  private def putMetric(paymentType: String) =
```
rupertbates: I think you could merge these two methods and just call toLowerCase() on the paymentType

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
rupertbates: I would remove CloudWatch from all the putX method names I think it's redundant

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -53,21 +47,35 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
...
 
-  def createPayPalPaymentMethod(payPal: PayPalPaymentFields, payPalService: PayPalService): Future[PayPalReferenceTransaction] =
+  def createPayPalPaymentMethod(payPal: PayPalPaymentFields, payPalService: PayPalService): Future[PayPalReferenceTransaction] = {
+    putCloudWatchCreatePaymentMethodMetric("paypal")
     payPalService
       .retrieveEmail(payPal.baid)
       .map(PayPalReferenceTransaction(payPal.baid, _))
+  }
+
+  private def putStartupMetric(paymentType: Either[StripePaymentFields, PayPalPaymentFields]) =
```
rupertbates: Which of the metrics is supposed to be fired first? Because currently the startup metric is the one that is being fired on success whereas the name sounds more like it is supposed to be on lambda startup and the other events would be on success

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
```diff
@@ -33,4 +35,14 @@ class SendThankYouEmail(thankYouEmailService: EmailService)
       product = "monthly-contribution"
     )).map(_ => Unit)
   }
+
+  private def putMetric(paymentType: String) =
```
rupertbates: As above - merge these two methods and just call toLowerCase() on the paymentType

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -53,21 +47,35 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
...
+    if (paymentType.isLeft)
+      putCloudWatchStartupMetric("stripe")
+    else
+      putCloudWatchStartupMetric("paypal")
+
+  def putCloudWatchStartupMetric(paymentMethod: String): Future[Unit] =
+    new RecurringContributionsMetrics(paymentMethod, "monthly")
+      .putContributionSignUpStartProcess().recover({ case _ => () })
```
rupertbates: I think this is supposed to call putCreatePaymentMethod() isn't it?

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreateSalesforceContact.scala
michaelwmcnamara: Good idea! Will fix in next commit.

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -53,22 +47,40 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
       state.acquisitionData
     )
 
-  def createStripePaymentMethod(stripe: StripePaymentFields, stripeService: StripeService): Future[CreditCardReferenceTransaction] =
+  def createStripePaymentMethod(stripe: StripePaymentFields, stripeService: StripeService): Future[CreditCardReferenceTransaction] = {
```
rupertbates: Don't need the brackets here now that the extra line has gone.

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -53,22 +47,40 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
...
       .createCustomer(stripe.userId, stripe.stripeToken)
       .map { stripeCustomer =>
         val card = stripeCustomer.card
         CreditCardReferenceTransaction(card.id, stripeCustomer.id, card.last4,
           CountryGroup.countryByCode(card.country), card.exp_month, card.exp_year, card.zuoraCardType)
       }
+  }
 
-  def createPayPalPaymentMethod(payPal: PayPalPaymentFields, payPalService: PayPalService): Future[PayPalReferenceTransaction] =
+  def createPayPalPaymentMethod(payPal: PayPalPaymentFields, payPalService: PayPalService): Future[PayPalReferenceTransaction] = {
```
rupertbates: Don't need the brackets here now that the extra line has gone.

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -53,22 +47,40 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
...
     payPalService
       .retrieveEmail(payPal.baid)
       .map(PayPalReferenceTransaction(payPal.baid, _))
+  }
+
+  private def putStartup(paymentType: Either[StripePaymentFields, PayPalPaymentFields]) =
```
rupertbates: You can refactor out this code duplication by replacing `putStartup` & `putPaymentMethodCreated` with:

      private def putMetric(paymentType: Either[StripePaymentFields, PayPalPaymentFields], f: String => Future[Unit]) =
        if (paymentType.isLeft)
          f("stripe")
        else
          f("paypal")

And then call this method above with `putMetric(state.paymentFields, startupMetric)` & `putMetric(state.paymentFields, paymentMethodCreatedMetric)`

michaelwmcnamara:
Latest commit should satisfy all requests from last review.

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
michaelwmcnamara: nice!

michaelwmcnamara:
@rupertbates - Latest commit has requested changes.

michaelwmcnamara:
This change was initially introduced as a workaround to the fact that Librato was unable to pull through all the standard Lambda metrics.
They have since managed to fix this bug and all tags for the standard AWS Lambda metrics are now pulling through so I can use those for alerting instead.
As a result I am reverting this change (no sense paying for something we already have).

PR #69, opened 2017-10-23
Upgrade workers and producer dependencies

desbo:
## Why are you doing this?
To use the latest Thrift definitions (via acquisition-event-producer)


desbo:
@rupertbates cheers - tested successfully on CODE

PR #70, opened 2017-10-24
set stripe payment metrics to have a better label

michaelwmcnamara:
Metrics tracking payments were added in https://github.com/guardian/support-workers/pull/68
These metrics are pulling through ok, however the payment method label for stripe payments is coming through as "CreditCardReferenceTransaction", not "stripe".

This change fixes that problem and will make the metrics easier to follow.

### DIFF
--- a/monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/ContributionCompleted.scala
+++ b/monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/ContributionCompleted.scala
@@ -24,7 +24,7 @@ class ContributionCompleted
     )
 
     logger.info(fields.map({ case (k, v) => s"$k: $v" }).mkString("SUCCESS ", " ", ""))
-    putContributionCompleted(state.paymentMethod.`type`)
+    putContributionCompleted(state.paymentMethod.toString)
 
     CompletedState(
       requestId = state.requestId,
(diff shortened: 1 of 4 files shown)


monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/ContributionCompleted.scala
```diff
@@ -35,8 +35,15 @@ class ContributionCompleted
     )
   }
 
-  def putContributionCompleted(paymentMethod: String): Future[Unit] =
-    new RecurringContributionsMetrics(paymentMethod.toLowerCase, "monthly")
+  def putContributionCompleted(paymentMethod: String): Future[Unit] = {
+    val paymentMethodLabel = {
```
rupertbates: A better way to do this (because it doesn't rely on string matches and will be caught by the compiler if the classname changes) is to use a match on the classtype:

      def putContributionCompleted(paymentMethod: PaymentMethod): Future[Unit] = {
        val paymentMethodLabel = paymentMethod match {
          case CreditCardReferenceTransaction => "stripe"
          case PaypalPaymentReferenceTransaction => "paypal"
        }
        new RecurringContributionsMetrics(paymentMethodLabel, "monthly")
          .putContributionCompleted().recover({ case _ => () })
      }

Also I think it would be good to refactor the RecurringContributionsMetrics class so that it takes a PaymentMethod object rather than a string, then you can do the conversion there (actually in the ProductDimensions class probably) rather than in the calling class and you'll only need to write code for it once.

michaelwmcnamara: That sounds like a much more elegant solution. Thanks!

PR #71, opened 2017-10-26
Revert "adds cloudwatch metrics for each lambda function"

michaelwmcnamara:
Reverts guardian/support-workers#68

This is reverting a change that added extra cloudwatch metrics to our Lambda functions.
This was so that we could take advantage of the composite metrics and alerting in Librato tocreate alerts on the rates of success vs attempts on out Lambda functions.

The new metrics were necessary because the full complement of tags for the standard AWS Lambda metrics were not being pulled through by Librato - as a result they were missing critical data.

Librato has just informed me that they have managed to fix this bug and I have tested and confirmed that all tags for the standard AWS Lambda metrics are now pulling through.
As a result the duplicate metrics are no longer needed and can be removed (no sense paying for something we already have).


jacobwinch:
👍

PR #72, opened 2017-10-31
Fix test JSON

jacobwinch:
## Why are you doing this?
To fix the tests.

## Changes
* Fix country value in mock JSON input 

cc @michaelwmcnamara


michaelwmcnamara:
😻

PR #73, opened 2017-11-14
Don't return a recurring contribution if the user has cancelled it

jacobwinch:
## Why are you doing this?

I noticed that the CreateZuoraSubscription lambda skips over the subscribe call for any user who has a subscription with a Contributor plan. This clause is too inclusive - it means that we are accidentally stopping users with cancelled contributions from signing up again.

[**Trello Card**](https://trello.com/c/p68imE2D/1092-allow-cancelled-monthly-contributors-to-re-subscribe)

## Changes

* Check that a sub is active when searching for a monthly contribution
* Add integration tests
* Minor logging/formatting tweaks


monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -27,7 +27,7 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
     }
 
   def skipSubscribe(state: CreateZuoraSubscriptionState, subscription: SubscriptionResponse): Future[SendThankYouEmailState] = {
-    logger.info(s"Skipping subscribe for user because they are already a contributor " +
+    logger.info(s"Skipping subscribe for user ${state.user.id} because they are already an active contributor " +
```
rupertbates: I think we want to avoid writing user ids into logs, just from a GDPR paranoia point of view.

jacobwinch: Sure, I'll remove this - although I do wonder how we are going to address this problem more generally (and also exactly what the status of account number will be in a GDPR compliant world!)

jacobwinch:
This is working on CODE.

PR #76, opened 2017-11-15
Use latest riff-raff plugin

jacobwinch:
## Why are you doing this?

1. To keep things up to date 
2. Because this version of the plugin seems to ensure that tests are passing before uploading to S3, and we had a problem before where riff-raff would deploy a build, even if the tests on master were red - so this gives us some additional safety.

## Changes

* Update plugin

I can deploy this branch to CODE, and I'm assuming that's the only testing required for this.


jacobwinch:
(This has already been done for support-frontend - see https://github.com/guardian/support-frontend/pull/311/files#diff-a84f91f20f040218bccd09fed4761fb3R15)

PR #77, opened 2017-11-23
Fail on exceptions

rupertbates:
## Why are you doing this?
To make it easier to see the outcome of state machine executions

This PR introduces 2 main changes 

* New RequestInfo info included with each request - this is unencrypted and contains useful information about the request such as whether the user is a test user and a list of messages which individual lambdas can add log messages to to make it easier to associate debug information with particular executions.
* New Failed state - previously the FailureHandler state would catch and swallow all exceptions meaning firstly that it was difficult to spot failed executions because they would still show as succeeded in the AWS console and secondly that routine failures caused by card payment failures were treated in the same way as real exceptions. Now card payment failures will continue to show as succeeded in the AWS console but any other error will transfer to a new Failed terminal state.

__Visual representation of an execution with a real error:__
screen shot 2017-11-23 at 12 49 55
__Visual representation of an execution with payment failure:__
screen shot 2017-11-23 at 12 50 56
__The structure of the new RequestInfo:__
screen shot 2017-11-23 at 12 51 31

[**Trello Card**](https://trello.com/c/woFAILTI/1090-make-it-easier-to-spot-failures-in-the-monthly-contributions-step-function-executions)

### DIFF
--- a/cloud-formation/src/templates/state-machine.yaml
+++ b/cloud-formation/src/templates/state-machine.yaml
@@ -54,6 +54,19 @@ States:
           {{> retry}}
     Next: WaitForBrowserPoll
 
+  FailureHandler:
+    Type: Task
+    Resource: "${FailureHandlerLambda.Arn}"
+    Next: SucceedOrFailChoice
+
+  SucceedOrFailChoice:
+    Type: Choice
+    Choices:
+      - Variable: "$.status"
+        StringEquals: "exception"
+        Next: FailState
+    Default: WaitForBrowserPoll
+
   WaitForBrowserPoll:
     Type: Wait
     Seconds: 30
@@ -63,7 +76,6 @@ States:
       is created and the old is terminated.
     End: True
 
-  FailureHandler:
-    Type: Task
-    Resource: "${FailureHandlerLambda.Arn}"
-    Next: WaitForBrowserPoll
+  FailState:
+    Type: Fail
+
--- a/cloud-formation/src/templates/state-machine.yaml
+++ b/cloud-formation/src/templates/state-machine.yaml
@@ -62,8 +62,8 @@ States:
   SucceedOrFailChoice:
     Type: Choice
     Choices:
-      - Variable: "$.status"
-        StringEquals: "exception"
+      - Variable: "$.requestInfo.failed"
+        BooleanEquals: true
         Next: FailState
     Default: WaitForBrowserPoll
 
--- a/common/src/main/scala/com/gu/zuora/encoding/CustomCodecs.scala
+++ b/common/src/main/scala/com/gu/zuora/encoding/CustomCodecs.scala
@@ -69,9 +69,14 @@ trait ModelsCodecs {
   implicit val encodeContribution: Encoder[Contribution] = deriveEncoder
 
   implicit val executionErrorCodec: Codec[ExecutionError] = deriveCodec
+
+  implicit val requestInformationCodec: Codec[RequestInformation] = deriveCodec
+
   implicit val jsonWrapperDecoder: Decoder[JsonWrapper] = Decoder
-    .forProduct3("state", "error", "encrypted")(JsonWrapper.apply)
-    .or(Decoder.forProduct2("state", "error")((s: String, e: Option[ExecutionError]) => JsonWrapper(s, e, encrypted = false)))
+    .forProduct3("state", "error", "requestInformation")(JsonWrapper.apply)
+    .or(Decoder.forProduct3("state", "error", "encrypted")((s: String, e: Option[ExecutionError], en: Boolean) =>
+        JsonWrapper(s, e, RequestInformation(en, testUser = false, failed = false, Nil))))
+
   implicit val jsonWrapperEncoder: Encoder[JsonWrapper] = deriveEncoder
 
   implicit val acquisitionDataCodec: Codec[AcquisitionData] = {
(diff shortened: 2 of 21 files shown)


monthly-contributions/src/test/scala/com/gu/support/workers/errors/PayPalErrorsSpec.scala
```diff
@@ -1,46 +1,41 @@
...
-    val outStream = new ByteArrayOutputStream()
-
-    a[RetryUnlimited] should be thrownBy {
-      createPaymentMethod.handleRequest(wrapFixture(createPayPalPaymentMethodJson()), outStream, context)
-    }
-
-    // Shut down the server. Instances cannot be reused.
-    server.shutdown()
-  }
+  //  "Timeouts from PayPal" should "throw a NonFatalException" in {
```
jacobwinch: Is this intentionally left commented out?

rupertbates: Nope, well spotted!

monthly-contributions/src/test/scala/com/gu/support/workers/Fixtures.scala
```diff
@@ -137,21 +138,30 @@ object Fixtures {
...
+          "error": {
+            "Error": "com.example.SomeError",
+            "Cause": "Oh no! It's on fire!"
```
jacobwinch: :fire:

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -20,23 +21,27 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
+  ): FutureHandlerResult =
     services.zuoraService.getRecurringSubscription(state.user.id, state.contribution.billingPeriod).flatMap {
-      case Some(sub) => skipSubscribe(state, sub)
-      case None => subscribe(state, services)
+      case Some(sub) => skipSubscribe(state, requestInfo, sub)
+      case None => subscribe(state, requestInfo, services)
     }
 
-  def skipSubscribe(state: CreateZuoraSubscriptionState, subscription: SubscriptionResponse): Future[SendThankYouEmailState] = {
-    logger.info(s"Skipping subscribe for user because they are already an active contributor " +
```
jacobwinch: Have we lost this logging, or does it happen in appendMessage?

rupertbates: I left it out, now that the message will appear in the console it didn't seem that useful.

cloud-formation/src/templates/state-machine.yaml
```diff
@@ -63,7 +76,6 @@ States:
       is created and the old is terminated.
     End: True
 
-  FailureHandler:
-    Type: Task
-    Resource: "${FailureHandlerLambda.Arn}"
-    Next: WaitForBrowserPoll
\ No newline at end of file
+  FailState:
```
jacobwinch: So presumably this causes the overall State Machine Execution status to be failed, which means that the status endpoint in support-frontend ([here?](https://github.com/guardian/support-frontend/blob/5d8b40e37cfb1509583b2fbd9b82ced3f6d468fc/app/services/stepfunctions/RegularContributionsClient.scala#L120)) will return a failure status?

rupertbates: Yes

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
jacobwinch: I'd prefer to keep it in the logs as well, if possible. It's useful to keep track of how often this is happening as it could be symptomatic of a wider problem.

I was actually thinking we might even want to add a metric or something around this. Although it's valuable as a safety check, theoretically we could break this check and stop taking money altogether, and we'd struggle to notice because everything would stay green!

common/src/main/scala/com/gu/zuora/encoding/CustomCodecs.scala
```diff
@@ -69,9 +69,14 @@ trait ModelsCodecs {
...
   implicit val executionErrorCodec: Codec[ExecutionError] = deriveCodec
+
+  implicit val RequestInfoCodec: Codec[RequestInfo] = deriveCodec
+
   implicit val jsonWrapperDecoder: Decoder[JsonWrapper] = Decoder
-    .forProduct3("state", "error", "encrypted")(JsonWrapper.apply)
-    .or(Decoder.forProduct2("state", "error")((s: String, e: Option[ExecutionError]) => JsonWrapper(s, e, encrypted = false)))
+    .forProduct3("state", "error", "requestInfo")(JsonWrapper.apply)
+    .or(Decoder.forProduct3("state", "error", "encrypted")((s: String, e: Option[ExecutionError], en: Boolean) =>
+      JsonWrapper(s, e, RequestInfo(en, testUser = false, failed = false, Nil))))
```
jacobwinch: Is this for backwards compatibility - tested here? https://github.com/guardian/support-workers/pull/77/files#diff-19f847463e299e4a039b9badecbc343dR26

rupertbates: Yes, it can go once this PR: https://github.com/guardian/support-frontend/pull/390 is deployed

PR #79, opened 2017-11-24
Add alarms for no executions and failed executions

jacobwinch:
## Why are you doing this?

This PR adds two new alarms, so that we know if things are broken in production:

**1. NoStateMachineExecutionsAlarm**
Intended to catch cases where the interaction between support-frontend and support-workers is fundamentally broken.

**2. ExecutionFailureAlarm** 
If the Fail State added in https://github.com/guardian/support-workers/pull/77 is triggered, we should be informed so that we can fix the bug (or add more genuine/valid reasons for failure to the handler).
If we break any of the individual lambdas, this should serve as a catch all for that too.

[**Trello Card**](https://trello.com/c/cnYz9x3R/1115-add-basic-production-monitoring-to-support-projects)

## Changes

* Add new alarms to Cloudformation

cc @michaelwmcnamara

### DIFF
--- a/cloud-formation/src/templates/cfn-template.yaml
+++ b/cloud-formation/src/templates/cfn-template.yaml
@@ -124,42 +124,42 @@ Resources:
 
   NoStateMachineExecutionsAlarm:
     Type: AWS::CloudWatch::Alarm
-    Condition: CreateCodeResources
+    Condition: CreateProdResources
     Properties:
       InsufficientDataActions:
-      - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:jacob-testing
+      - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:simple-dev
       OKActions:
-      - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:jacob-testing
+      - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:simple-dev
       AlarmName: !Sub No Recent State Machine Execution Attempts in ${Stage} (Monthly Contributions Sign-Ups)
       AlarmDescription: The Monthly Contribtions state machine has not been triggered for 3 hours - has something broken?
       MetricName: ExecutionsStarted
       Namespace: AWS/States
       Dimensions:
         - Name: StateMachineArn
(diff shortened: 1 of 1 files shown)


cloud-formation/src/templates/cfn-template.yaml
```diff
@@ -120,4 +120,46 @@ Resources:
             {{> stateMachine}}
 
           - {}
-      RoleArn: !GetAtt [ StatesExecutionRole, Arn ]
\ No newline at end of file
+      RoleArn: !GetAtt [ StatesExecutionRole, Arn ]
+
+  NoStateMachineExecutionsAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Condition: CreateProdResources
```
jacobwinch: We only create these for the PROD stack

cloud-formation/src/templates/cfn-template.yaml
```diff
@@ -120,4 +120,46 @@ Resources:
...
+    Properties:
+      InsufficientDataActions:
+      - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:simple-dev
+      OKActions:
```
jacobwinch: If this resolves itself (i.e. people start signing up again after a quiet period), we get emailed about that too

cloud-formation/src/templates/cfn-template.yaml
```diff
@@ -120,4 +120,46 @@ Resources:
...
+      AlarmName: !Sub No Recent State Machine Execution Attempts in ${Stage} (Monthly Contributions Sign-Ups)
+      AlarmDescription: The Monthly Contributions state machine has not been triggered for an extended period - has something broken?
+      MetricName: ExecutionsStarted
+      Namespace: AWS/States
+      Dimensions:
+        - Name: StateMachineArn
+          Value: !Ref MonthlyContributionsPROD
+      ComparisonOperator: LessThanThreshold
+      Threshold: 1
+      Period: 7200
```
jacobwinch: This might need some tuning, once I can work out exactly how AWS uses the period to determine whether there is insufficient data or not...

jacobwinch:
This has been tested in CODE (including deployment via riff-raff).

PR #80, opened 2017-11-29
Rework error handling

rupertbates:
## Why are you doing this?

Up until now we have been handling payment failures that occur when Zuora tries to charge a card, but not failures which come directly from Stripe as soon as we try to create the customer. 

I'm not sure why these occur because they __should__ be picked up by the Stripe UI, but they do so we need to handle them in the same way, ie. log the error and mark the execution as succeeded (we only mark executions as failed if it is an unexpected exception).

I've also taken the opportunity to tidy up some of the error handling code to make it a bit more usable and added some tests.

[**Trello Card**](https://trello.com/c/HgCxnIFG/1128-handle-payment-failures-directly-from-stripe)


JustinPinner:
Looks sensible to me. I wouldn't be offended if you sought a second opinion though :) 👍

jacobwinch:
Looks good to me 👍

PR #81, opened 2017-11-30
Add timeout to state machine and fire an alert if an execution times out

jacobwinch:
## Why are you doing this?

We've seen a couple of executions which have got 'stuck' in a Running state (i.e. they've been using RetryUnlimited for a task which will never succeed for the given input e.g. [people with dodgy email addresses in Identity](https://trello.com/c/GomdbCeN/957-the-salesforce-mystery-freaks-out-for-invalid-emails)).

This PR adds a shorter timeout / time limit to our state machine executions, and also adds an alert which triggers if this timeout is ever reached.

Originally we attempted to catch this problem using the [ExecutionTime metric](http://docs.aws.amazon.com/step-functions/latest/dg/procedure-cw-metrics.html#monitoring-using-cloudwatch-state-machine-metrics), but since this is calculated by subtracting the start time of the execution from the finish time, the metric doesn't actually receive any data points until the state machine execution stops. Without a timeout configured it looks like this will use the [default limit of 1 year](http://docs.aws.amazon.com/step-functions/latest/dg/limits.html#service-limits-state-machine-executions), and therefore won't be a particularly helpful metric for the cases that we're trying to catch here.

Any opinions on the length of the timeout (or whether adding a timeout at all is sensible)?  @rupertbates @davidfurey @michaelwmcnamara 

[**Trello Card**](https://trello.com/c/FBRME4O1/961-find-a-way-to-detect-monthly-contribution-sign-ups-that-are-stuck)

## Changes

1. Add 24 hour timeout to state machine executions
2. Add alert so we know if 1 happens


rupertbates:
I think this is definitely a good idea but I would make the timeout quite a bit longer as it's fairly likely we could get an outage longer than this. Maybe put it to a day at first and then we can tweak it later?

jacobwinch:
Thanks @rupertbates  - sounds sensible, I've tweaked the timeout and the PR description

PR #84, opened 2017-12-06
Send a Stripe-Version header if configured and expect the new Customer model from Stripe

lmath:
## Why are you doing this?
Goal: Be able to test out a newer version of Stripe app by app to remove the element of blind courage currently required to update the version in the Stripe dashboard.

[The Stripe Upgrade Guide](https://stripe.com/docs/upgrades) specifies that the API version used for a given stripe account is the one configured in the dashboard, unless the request has a Stripe-Version header. So, the plan is to send a header Stripe-Version -> 2017-08-15 for the GNM membership account in each relevant app and then up the version in the dashboard.

This change makes it possible to add a stripe version to the config that will be sent on each request to Stripe. 

Additionally, since version [2015-02-18](https://stripe.com/docs/upgrades#2015-02-18), the call to create a customer returns not a list of `cards` but `sources` so I have updated the response we are expecting from Stripe to the one corresponding to version 2017-08-15

Pre merge:
- [x] Test in CODE 
- [x] ensure stripe.api.version = "2017-08-15" is in the config for both CODE and PROD

[**Trello Card**](https://trello.com/c/OZ8I3TEN/3-auto-update-card-details-when-card-is-replaced)

## Changes

* On requests to Stripe, we send a version header if there is a version in the Stripe config. 
* Expect the newer Customer response we expect from Stripe after version 2015-02-18

### DIFF
--- a/common/src/main/scala/com/gu/stripe/Stripe.scala
+++ b/common/src/main/scala/com/gu/stripe/Stripe.scala
@@ -48,24 +48,24 @@ object Stripe {
     implicit val codec: Codec[Card] = deriveCodec
   }
 
-  case class Card(id: String, `type`: String, last4: String, exp_month: Int, exp_year: Int, country: String) extends StripeObject {
-    val issuer = `type`.toLowerCase
+  case class Card(id: String, brand: String, last4: String, exp_month: Int, exp_year: Int, country: String) extends StripeObject {
+    val issuer = brand.toLowerCase
     // Zuora requires 'AmericanExpress' not 'American Express'
     // See CreditCardType at https://www.zuora.com/developer/api-reference/#operation/Object_POSTPaymentMethod
-    val zuoraCardType = `type`.replaceAll(" ", "")
+    val zuoraCardType = brand.replaceAll(" ", "")
   }
 
   object Customer {
     implicit val codec: Codec[Customer] = deriveCodec
   }
 
-  case class Customer(id: String, cards: StripeList[Card]) extends StripeObject {
+  case class Customer(id: String, sources: StripeList[Card]) extends StripeObject {
     // customers should always have a card
-    if (cards.total_count != 1) {
-      throw StripeError("internal", s"Customer $id has ${cards.total_count} cards, should have exactly one")
+    if (sources.total_count != 1) {
+      throw StripeError("internal", s"Customer $id has ${sources.total_count} cards, should have exactly one")
     }
 
-    val card = cards.data.head
+    val card = sources.data.head
   }
 
   object Source {
--- a/common/src/main/scala/com/gu/stripe/Stripe.scala
+++ b/common/src/main/scala/com/gu/stripe/Stripe.scala
@@ -44,11 +44,11 @@ object Stripe {
 
   case class StripeList[T](total_count: Int, data: Seq[T]) extends StripeObject
 
-  object Card {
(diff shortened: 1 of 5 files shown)


common/src/main/scala/com/gu/stripe/StripeService.scala
```diff
@@ -15,9 +15,18 @@ class StripeService(config: StripeConfig, client: FutureHttpClient, baseUrl: Str
   val wsUrl = baseUrl
   val httpClient: FutureHttpClient = client
 
-  override def wsPreExecute(req: Request.Builder): Request.Builder =
+  override def wsPreExecute(req: Request.Builder): Request.Builder = {
     req.addHeader("Authorization", s"Bearer ${config.secretKey}")
 
+    config.version match {
```
davidfurey: Since this is side-effect-y, I would use `foreach` instead of `match`.

```
req.addHeader("Authorization", s"Bearer ${config.secretKey}")

config.version foreach { version =>
  logger.info(s"Making a stripe call with version: $version")
  req.addHeader("Stripe-Version", version)
}

req
```

Also, there is no need for `{}` to separate `case` statements

common/src/main/scala/com/gu/stripe/Stripe.scala
```diff
@@ -48,24 +48,24 @@ object Stripe {
     implicit val codec: Codec[Card] = deriveCodec
   }
 
-  case class Card(id: String, `type`: String, last4: String, exp_month: Int, exp_year: Int, country: String) extends StripeObject {
-    val issuer = `type`.toLowerCase
+  case class Card(id: String, brand: String, last4: String, exp_month: Int, exp_year: Int, country: String) extends StripeObject {
```
rupertbates: Should we change this class to `Source` so that `customer.sources` becomes a `StripeList[Source]` do you think?

lmath: I figured keeping it Card was OK since the sources in practice are still cards but quite happy to call it Source like the Stripe API does.

However this reminds me -- there is an [existing case class Source](https://github.com/guardian/support-workers/blob/master/common/src/main/scala/com/gu/stripe/Stripe.scala#L71) and I could not figure out where it was being used (other than within case class Charge, which in turn I couldn't see where it was used). I think I should understand this before I can rename Card to Source. Can you point me in the right direction for that?

common/src/main/scala/com/gu/stripe/StripeService.scala
lmath: Ah, good point. I shall use foreach instead.

common/src/main/scala/com/gu/stripe/Stripe.scala
rupertbates: Yes it looks as though you could get rid of those 2 classes, I've just tried it and it seems ok. There are a few failing tests but they fail on master as well, I think it's because of the Salesforce work and a data clearout that Jacob did from Zuora.

lmath: Thanks for checking that. I'm going to rename Card to Source then.

jacobwinch: @rupertbates @lmath re the failing tests - yes, that's why they're broken. I'm working with Paul to re-configure Zuora UAT. When that's done I'll create some new stub data and fix the tests.

PR #89, opened 2018-01-15
Retry on UNKNOWN_ERROR from Zuora

jacobwinch:
## Why are you doing this?

We were unable to sign-up new users during a recent [Zuora incident](https://trust.zuora.com/incidents/0yd5tp9k6z10). This PR allows us to cope with the specific problem that we saw (by retrying), should it happen again. 

There are probably a number of other cases that we should consider handling too, but at the moment I can't find any documentation which clearly details all of the expected errors. I'll open a support ticket to ask about this and then open new PRs if necessary.

[**Trello Card**](https://trello.com/c/lSpMrPhN/1211-add-retries-for-known-zuora-errors-which-occur-during-the-createzuorasubscription-task)

## Changes

* Add pattern match to catch 'UNKNOWN_ERROR', and convert to a RetryUnlimited.
* Add Zuora-specific test cases to ErrorHandlerSpec.

### DIFF
--- a/common/src/main/scala/com/gu/zuora/model/response/Responses.scala
+++ b/common/src/main/scala/com/gu/zuora/model/response/Responses.scala
@@ -3,7 +3,7 @@ package com.gu.zuora.model.response
 import cats.implicits._
 import com.gu.support.workers.encoding.Helpers.{capitalizingCodec, deriveCodec}
 import com.gu.support.workers.encoding.{Codec, ErrorJson}
-import com.gu.support.workers.exceptions.{RetryException, RetryNone}
+import com.gu.support.workers.exceptions.{RetryException, RetryNone, RetryUnlimited}
 import io.circe.parser._
 import io.circe.syntax._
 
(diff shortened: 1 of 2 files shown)


common/src/main/scala/com/gu/zuora/model/response/Responses.scala
```diff
@@ -34,7 +34,10 @@ object ZuoraErrorResponse {
 case class ZuoraErrorResponse(success: Boolean, errors: List[ZuoraError])
     extends Throwable(errors.asJson.spaces2) with ZuoraResponse {
 
-  def asRetryException: RetryException = new RetryNone(message = this.asJson.noSpaces, cause = this)
+  def asRetryException: RetryException = errors match {
+    case List(ZuoraError("UNKNOWN_ERROR", _)) => new RetryUnlimited(this.asJson.noSpaces, cause = this)
```
jacobwinch: Although this is technically a list I've never seen it contain more than one item...

rupertbates: Is this definitely the form that the error deserialises to?

jacobwinch: Yep I'm pretty sure it does - there's an example in the logs [here](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#logEventViewer:group=/aws/lambda/support-CreateZuoraSubscriptionLambda-PROD;filter=UNKNOWN_ERROR;start=2018-01-11T18:17:00Z;end=2018-01-11T18:19:00Z).

PR #95, opened 2018-02-01
New Stripe config

rupertbates:
## Why are you doing this?
In order support payments through Stripe in Australian dollars we need to use a different account and Zuora Payment Gateway.

### Changes
* Use new version of support-config to allow us to have different Stripe configs for Australia and the rest of the world.
* Use a different payment gateway for Stripe depending on the currency

### DIFF
--- a/common/src/main/scala/com/gu/helpers/WebServiceHelper.scala
+++ b/common/src/main/scala/com/gu/helpers/WebServiceHelper.scala
@@ -7,10 +7,10 @@ import io.circe.parser._
 import io.circe.{Decoder, Json, Printer}
 import okhttp3._
 
+import scala.collection.immutable.Map.empty
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 import scala.reflect.{ClassTag, classTag}
-import scala.collection.JavaConversions._
 
 case class WebServiceHelperError[T: ClassTag](responseCode: Int, responseBody: String) extends Throwable {
   override def getMessage: String = s"${classTag[T]} - $responseCode: $responseBody"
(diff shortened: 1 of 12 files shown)


common/src/main/scala/com/gu/zuora/model/PaymentGateway.scala
```diff
@@ -11,14 +13,15 @@ sealed trait PaymentGateway {
 }
 
 object PaymentGateway {
-  def forPaymentMethod(paymentMethod: PaymentMethod): PaymentGateway = paymentMethod match {
+  def forPaymentMethod(paymentMethod: PaymentMethod, currency: Currency): PaymentGateway = paymentMethod match {
     case _: PayPalReferenceTransaction => PayPalGateway
-    case _: CreditCardReferenceTransaction => StripeGateway
+    case _: CreditCardReferenceTransaction => if (currency == AUD) StripeGatewayAUD else StripeGatewayDefault
```
davidfurey: I'd have a slight preference to split this into two cases:
```
case _: CreditCardReferenceTransaction if (currency == AUD) => StripeGatewayAUD
case _: CreditCardReferenceTransaction => StripeGatewayDefault

File: common/src/main/scala/com/gu/helpers/WebServiceHelper.scala [hunk 2]
```diff
@@ -66,32 +68,40 @@ trait WebServiceHelper[Error <: Throwable] extends LazyLogging {
...
    * @param responseBody
    * @param errorDecoder
    * @return Either[circe.Error, Error]
    */
   def decodeError(responseBody: String)(implicit errorDecoder: Decoder[Error]): Either[circe.Error, Error] = decode[Error](responseBody)
 
-  def get[A](endpoint: String, params: (String, String)*)(implicit decoder: Decoder[A], errorDecoder: Decoder[Error], ctag: ClassTag[A]): Future[A] =
-    request[A](new Request.Builder().url(endpointUrl(endpoint, params)))
+  //Scalariform won't let you add a line break in method signatures, but then scalaStyle freaks out because they're too long
+  // scalastyle:off line.size.limit
```
davidfurey: It would be nice if Scalariform and ScalaStyle wouldn't argue so much

rupertbates:
@davidfurey we still have wsPreExecute there because the SalesforceService needs to manipulate the url before sending and changing that would be more complicated, but I've switched everything else over to passing the headers in up front.

PR #96, opened 2018-02-01
Improve Zuora error handling

jacobwinch:
## Why are you doing this?

This is a follow up to https://github.com/guardian/support-workers/pull/89. This should allow us to cope with all retry-able errors that we might receive from Zuora.

[**Trello Card**](https://trello.com/c/lSpMrPhN/1211-add-retries-for-zuora-errors-which-occur-during-the-createzuorasubscription-task)

## Changes

* Add more known errors to the asRetryException pattern match (and test them)
* Update error handling docs

### DIFF
--- a/common/src/main/scala/com/gu/zuora/model/response/Responses.scala
+++ b/common/src/main/scala/com/gu/zuora/model/response/Responses.scala
@@ -34,12 +34,21 @@ object ZuoraErrorResponse {
 case class ZuoraErrorResponse(success: Boolean, errors: List[ZuoraError])
     extends Throwable(errors.asJson.spaces2) with ZuoraResponse {
 
+  override def toString: String = this.errors.toString()
+  def toRetryNone: RetryNone = new RetryNone(message = this.asJson.noSpaces, cause = this)
+  def toRetryUnlimited: RetryUnlimited = new RetryUnlimited(this.asJson.noSpaces, cause = this)
+
+  // Based on https://knowledgecenter.zuora.com/DC_Developers/G_SOAP_API/L_Error_Handling/Errors#ErrorCode_Object
   def asRetryException: RetryException = errors match {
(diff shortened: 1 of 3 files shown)


common/src/main/scala/com/gu/zuora/model/response/Responses.scala
```diff
@@ -34,12 +34,21 @@ object ZuoraErrorResponse {
...
 
+  override def toString: String = this.errors.toString()
+  def toRetryNone: RetryNone = new RetryNone(message = this.asJson.noSpaces, cause = this)
+  def toRetryUnlimited: RetryUnlimited = new RetryUnlimited(this.asJson.noSpaces, cause = this)
+
+  // Based on https://knowledgecenter.zuora.com/DC_Developers/G_SOAP_API/L_Error_Handling/Errors#ErrorCode_Object
   def asRetryException: RetryException = errors match {
-    case List(ZuoraError("UNKNOWN_ERROR", _)) => new RetryUnlimited(this.asJson.noSpaces, cause = this)
-    case _ => new RetryNone(message = this.asJson.noSpaces, cause = this)
+    case List(ZuoraError("API_DISABLED", _)) => toRetryUnlimited
```
jacobwinch: Hopefully this never happens, but if it does we'll now have 24 hours to fix it before we lose the contributors!

rupertbates: All of these will only match against a list with one item (as with the original code). I think this is a reasonable assumption based on what we've seen so far but I guess we may need to revisit at some point

jacobwinch: Yep, I tried to explicitly acknowledge this assumption in the error handling docs. Unfortunately it does feel like there's going to be a certain amount of trial and error involved with getting the Zuora error handling completely right...

PR #101, opened 2018-02-20
Upgrade support-internationalisation

svillafe:
## Why are you doing this?

To keep support-internationalisation as up to date as possible.

## Changes

* `Dependencies.scala`


svillafe:
Tested in CODE and it is working.

PR #103, opened 2018-02-21
Add retries to FailureHandler task

jacobwinch:
## Why are you doing this?

This should help to ensure that users who fail to sign-up always receive an email. It should also cut down on the noise from our AWS alerts, as currently if this lambda fails we get an alert due to the Step Function execution failing.

I've set up some temporary alerting on the two lambdas which are timing out, so that I can continue debugging this without spamming a huge mailing list. (Full details on approach can be found in the Trello card).

[**Trello Card**](https://trello.com/c/Fh3vvzo7/1296-add-retries-to-cope-with-cases-where-failurehandler-throws-lambdaunknown-errors)

## Changes

* Use the standard [retry rules](https://github.com/guardian/support-workers/blob/master/cloud-formation/src/templates/retries.yaml) for the FailureHandler.

### DIFF
--- a/cloud-formation/src/templates/state-machine.yaml
+++ b/cloud-formation/src/templates/state-machine.yaml
@@ -52,6 +52,7 @@ States:
     Type: Task
     Resource: "${FailureHandlerLambda.Arn}"
     Next: SucceedOrFailChoice
+    {{> retry}}
 
   SucceedOrFailChoice:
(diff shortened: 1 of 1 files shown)


jacobwinch:
Decided that the standard retry rules are not safe/sensible for lambdas which send emails, after discussing with @rupertbates.

PR #105, opened 2018-02-22
Add more logging to the email service

jacobwinch:
## Why are you doing this?
This is a follow up to https://github.com/guardian/support-workers/pull/102.

[**Trello Card**](https://trello.com/c/Fh3vvzo7/1296-add-retries-to-cope-with-cases-where-failurehandler-throws-lambdaunknown-errors)

## Changes
* Log the result when sending a message to SQS

### DIFF
--- a/common/src/main/scala/com/gu/emailservices/EmailService.scala
+++ b/common/src/main/scala/com/gu/emailservices/EmailService.scala
@@ -6,8 +6,8 @@ import com.amazonaws.services.sqs.model.{SendMessageRequest, SendMessageResult}
 import com.gu.aws.{AwsAsync, CredentialsProvider}
 import com.typesafe.scalalogging.StrictLogging
 import org.joda.time.DateTime
-
 import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits.global
 import scala.util.{Failure, Success}
 
 case class EmailFields(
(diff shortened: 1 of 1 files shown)


common/src/main/scala/com/gu/emailservices/EmailService.scala
```diff
@@ -6,8 +6,9 @@ import com.amazonaws.services.sqs.model.{SendMessageRequest, SendMessageResult}
 import com.gu.aws.{AwsAsync, CredentialsProvider}
 import com.typesafe.scalalogging.StrictLogging
 import org.joda.time.DateTime
-
 import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits.global
```
hosamaly: If I may make a suggestion, I think that it's better to let the `send` method accept an `ExecutionContext`, giving the caller control over which thread pool to use.

hosamaly: Alternatively, it's possible to use a same-thread execution context (such as [Guava's `directExecutor`](https://google.github.io/guava/releases/24.0-jre/api/docs/com/google/common/util/concurrent/MoreExecutors.html#directExecutor--)) to avoid the need for a thread pool.

jacobwinch: Thanks for the suggestions. In the end I decided to allow clients to pass in an execution context when creating an instance of [EmailService](https://github.com/guardian/support-workers/pull/106/files#diff-dccf2741170d67c2b8991c9bf18960af).

PR #106, opened 2018-02-23
Add even more logging to help understand task timeouts

jacobwinch:
## Why are you doing this?

Because I still can't debug the email task timeouts...

[**Trello Card**](https://trello.com/c/Fh3vvzo7/1296-stop-failurehandler-from-throwing-lambdaunknown-errors)

## Changes
* Add logging to AWSAsyncHandler, so we can ascertain whether we get a response from AWS
* Move from onComplete to recover/map, to ensure that the email service logging completes before the lambda exits
* EmailService class now accepts an execution context

### DIFF
--- a/common/src/main/scala/com/gu/aws/AWSAsyncHandler.scala
+++ b/common/src/main/scala/com/gu/aws/AWSAsyncHandler.scala
@@ -14,19 +14,25 @@ class AwsAsyncHandler[Request <: AmazonWebServiceRequest, Response](f: (Request,
 
   private val promise = Promise[Response]()
 
-  override def onError(exception: Exception): Unit = exception match {
-    case e: AmazonServiceException =>
-      if (e.getErrorCode == "ThrottlingException") {
-        logger.warn(s"A rate limiting exception was thrown, we may need to adjust the rate limiting in ClientWrapper.scala")
-        Thread.sleep(1000) //Wait for a second and retry
-        f(request, this)
-      } else {
-        promise.failure(exception)
-      }
-    case _ => promise.failure(exception)
+  override def onError(exception: Exception): Unit = {
+    logger.warn(s"Failure from AWSAsyncHandler", exception)
+    exception match {
+      case e: AmazonServiceException =>
+        if (e.getErrorCode == "ThrottlingException") {
+          logger.warn(s"A rate limiting exception was thrown, we may need to adjust the rate limiting in ClientWrapper.scala")
+          Thread.sleep(1000) //Wait for a second and retry
+          f(request, this)
+        } else {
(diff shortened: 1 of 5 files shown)


common/src/main/scala/com/gu/aws/AWSAsyncHandler.scala
```diff
@@ -14,19 +14,25 @@ class AwsAsyncHandler[Request <: AmazonWebServiceRequest, Response](f: (Request,
...
-      if (e.getErrorCode == "ThrottlingException") {
-        logger.warn(s"A rate limiting exception was thrown, we may need to adjust the rate limiting in ClientWrapper.scala")
-        Thread.sleep(1000) //Wait for a second and retry
-        f(request, this)
-      } else {
-        promise.failure(exception)
-      }
-    case _ => promise.failure(exception)
+  override def onError(exception: Exception): Unit = {
+    logger.warn(s"Failure from AWSAsyncHandler", exception)
```
davidfurey: No need for `s""` if you are doing string interpolation

common/src/main/scala/com/gu/aws/AWSAsyncHandler.scala
```diff
@@ -14,19 +14,25 @@ class AwsAsyncHandler[Request <: AmazonWebServiceRequest, Response](f: (Request,
...
+        } else {
+          promise.failure(exception)
+        }
+      case _ => promise.failure(exception)
+    }
   }
 
-  override def onSuccess(request: Request, result: Response): Unit = promise.success(result)
+  override def onSuccess(request: Request, result: Response): Unit = {
+    logger.info(s"Successful result from AWS AsyncHandler $result")
```
davidfurey: I think there is a good change that `$result` will print the java hashcode for the object, since it isn't a nice Scala case class.  But no harm in including it

common/src/main/scala/com/gu/emailservices/EmailService.scala
```diff
@@ -42,7 +40,8 @@ case class EmailFields(
     """.stripMargin
 }
 
-class EmailService(config: EmailConfig) extends StrictLogging {
+class EmailService(config: EmailConfig, executionContext: ExecutionContext) extends StrictLogging {
```
davidfurey: You should be able to write:
`class EmailService(config: EmailConfig, implicit val executionContext: ExecutionContext)`
to avoid needing `ec` further down

common/src/main/scala/com/gu/aws/AWSAsyncHandler.scala
jacobwinch: Thanks, good spot

PR #107, opened 2018-03-09
Delete glitchy alarm

jacobwinch:
## Why are you doing this?

This alarm is not working as intended and is adding 0 value (see recent spam emails). 

I'd like to delete it whilst I think about better approaches to monitoring for this class of problem.

## Changes

* Remove alert from Cloudformation

### DIFF
--- a/cloud-formation/src/templates/cfn-template.yaml
+++ b/cloud-formation/src/templates/cfn-template.yaml
@@ -122,28 +122,6 @@ Resources:
           - {}
       RoleArn: !GetAtt [ StatesExecutionRole, Arn ]
 
-  NoStateMachineExecutionsAlarm:
-    Type: AWS::CloudWatch::Alarm
-    Condition: CreateProdResources
-    Properties:
-      InsufficientDataActions:
-      - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
-      OKActions:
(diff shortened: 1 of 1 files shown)


cloud-formation/src/templates/cfn-template.yaml
```diff
@@ -122,28 +122,6 @@ Resources:
...
       RoleArn: !GetAtt [ StatesExecutionRole, Arn ]
 
-  NoStateMachineExecutionsAlarm:
-    Type: AWS::CloudWatch::Alarm
-    Condition: CreateProdResources
-    Properties:
-      InsufficientDataActions:
-      - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
-      OKActions:
-      - !Sub arn:aws:sns:${AWS::Region}:${AWS::AccountId}:reader-revenue-dev
```
johnduffell: surely the glitch is here - we want AlarmActions not the above two?

jacobwinch:
The point of this alert was to notify us if support-frontend was unable to call the Step Function (which signs up recurring contributors) for some reason. 

I thought that I could capture this fairly simply by alerting on InsufficientData, but in practice this doesn't seem to work particularly well. So yes @svillafe, I think we want to replace monitoring of this integration eventually, but perhaps we can't do it with a straight forward CloudWatch alert...

johnduffell:
undestood - @pvighi had the same issue with fulfilment alarms

PR #108, opened 2018-03-22
Relocate dependencies into build.sbt so snyk can get at them

JustinPinner:
## Why are you doing this?
Snyk doesn't seem to be able to unravel our dependencies when they're defined in `Dependencies.scala`. After a brief discussion with @rupertbates and @Ap0c last night, we concluded it may work if they're in `build.sbt`, so that's where I've put them to see how it goes.

[**Trello Card 1392**](https://trello.com/c/RSS0jyY7/1392-send-support-workers-dependencies-to-snyk-as-part-of-ci)

## Changes

* See that `Dependencies.scala`? That's gone that is.
* aws 1.11.131 -> 1.11.161
* dispatch-core 0.11.3 -> 0.11.4
* Ignore vulnerabilites from jackson-databind 2.6.7.1


JustinPinner:
Well, it seems that snyk has woken up!

rupertbates:
👍 Looks good, do you want to take on the card of fixing these errors? I was working on it but I think it makes sense to roll it into this PR.

JustinPinner:
@rupertbates Yep, I'll have a look at those things, it'll be good practice for me. Also, does this change match what you had in mind when we spoke about it yesterday?

rupertbates:
@JustinPinner yes it does.

JustinPinner:
Ok, good. Snyk is happy again. Time to test on CODE...

JustinPinner:
BTW the tests on CODE checked out all the way through to Zuora, so this should be good to go as soon as I can get an approved review.

PR #109, opened 2018-03-28
Configure thread pool for thank you email lambda

jacobwinch:
## Why are you doing this?

Since we're running in a lambda with relatively limited resources, the max number of threads available (when using the global execution context) is always likely to be pretty low. We think that the lambda timeouts that we're seeing may be due to thread contention, so this PR attempts to test that theory.

[**Trello Card**](https://trello.com/c/Fh3vvzo7/1296-stop-failurehandler-from-throwing-lambdaunknown-errors)

## Changes

* Configure new execution context with fixed thread pool
* Log the number of available processors to help with debugging (see https://docs.scala-lang.org/overviews/core/futures.html#the-global-execution-context)

### DIFF
--- a/common/src/main/scala/com/gu/threadpools/CustomPool.scala
+++ b/common/src/main/scala/com/gu/threadpools/CustomPool.scala
@@ -1,0 +1,11 @@
+package com.gu.threadpools
+
+import java.util.concurrent.Executors
+import com.typesafe.scalalogging.StrictLogging
+import scala.concurrent.ExecutionContext
+
+object CustomPool extends StrictLogging {
+
+  implicit val ec = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(4))
+
+}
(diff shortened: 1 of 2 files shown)


common/src/main/scala/com/gu/threadpools/CustomPool.scala
```diff
@@ -0,0 +1,11 @@
+package com.gu.threadpools
+
+import java.util.concurrent.Executors
+import com.typesafe.scalalogging.StrictLogging
+import scala.concurrent.ExecutionContext
+
+object CustomPool extends StrictLogging {
+
+  implicit val ec = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(4))
```
davidfurey: This doesn't need to be implicit, and could have a more meaningful name (`executionContext`)

common/src/main/scala/com/gu/threadpools/CustomPool.scala
```diff
@@ -0,0 +1,11 @@
+package com.gu.threadpools
+
+import java.util.concurrent.Executors
+import com.typesafe.scalalogging.StrictLogging
+import scala.concurrent.ExecutionContext
+
+object CustomPool extends StrictLogging {
```
davidfurey: If you aren't going to log anything, there is no need to extend `StrictLogging`

PR #110, opened 2018-03-28
Move log statement somewhere sensible

jacobwinch:
## Why are you doing this?

To ensure that we are collecting the debug information that we need / generally tidying up after myself.


michaelwmcnamara:
+1

PR #116, opened 2018-04-20
Wait for email result before allowing handler to exit

jacobwinch:
## Why are you doing this?

We are not currently waiting for the asynchronous call which puts an email message onto an SQS queue. This means that the handler can exit before we've confirmed that the result from AWS was successful.

Although we wait for a Future to complete [here](https://github.com/guardian/support-workers/blob/980394a0e8a2f112a30bb5069af1a160c6a8d694/common/src/main/scala/com/gu/support/workers/lambdas/Handler.scala#L46-L51), we have ended up with nested Futures and were discarding the one that we care about on line 33.

We will have failed to send thank you emails to some recurring contributors since this issue was introduced.


jacobwinch:
@davidfurey also suggested that we look at ways to guard against this problem in the future - I'll create a card for that but would prefer to get this fix out first.

jacobwinch:
The follow-up card is here: https://trello.com/c/IoLviTwE/1449-ensure-that-we-wait-for-asynchronous-calls-to-complete-before-allowing-lambda-executions-to-complete

PR #118, opened 2018-05-22
Stop sending PII to Sentry and use SafeLogger everywhere

lmath:
## Why are you doing this?
We want to stop sending PII to sentry as we don't typically use it to debug problems at a user level. More justification here: https://github.com/guardian/support-frontend/pull/494

[**Trello Card**](https://trello.com/c/kAuozREd/1501-strip-any-pii-from-support-workers-logs-errors-before-sending-them-to-sentry)

## Changes

* Brought in the SafeLogger from https://github.com/guardian/support-frontend/pull/494
* Used SafeLogger everywhere 
* Added an additional log filter, which removes any error messages that have not been scrubbed and applied it to the Sentry appender

### DIFF
--- a/common/src/main/scala/com/gu/monitoring/SafeLogger.scala
+++ b/common/src/main/scala/com/gu/monitoring/SafeLogger.scala
@@ -1,0 +1,46 @@
+package com.gu.monitoring
+
+import com.typesafe.scalalogging.StrictLogging
+import org.slf4j.{Marker, MarkerFactory}
+
+object SafeLogger extends StrictLogging {
+
+  val sanitizedLogMessage: Marker = MarkerFactory.getMarker("SENTRY")
+
+  case class LogMessage(withPersonalData: String, withoutPersonalData: String) {
+    override val toString = withoutPersonalData
+  }
+
+  implicit class Sanitizer(val sc: StringContext) extends AnyVal {
+    def scrub(args: Any*): LogMessage = {
+      LogMessage(sc.s(args: _*), sc.s(args.map(_ => "*****"): _*))
+    }
+  }
+
+  def debug(logMessage: String): Unit = {
+    logger.debug(logMessage)
+  }
+
+  def info(logMessage: String): Unit = {
+    logger.info(logMessage)
+  }
+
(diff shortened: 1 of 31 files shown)


common/src/main/scala/com/gu/monitoring/SafeLogger.scala
```diff
@@ -0,0 +1,46 @@
+package com.gu.monitoring
```
lmath: I actually just copy/pasted SafeLogger from membership-common but it's also used in support-frontend: https://github.com/guardian/support-frontend/blob/master/app/monitoring/Logging.scala 

Is there some sensible place for this to live where it can be used by both or shall I just keep this here?

monthly-contributions/src/main/scala/com/gu/support/workers/lambdas/CreatePaymentMethod.scala
```diff
@@ -21,7 +23,8 @@ class CreatePaymentMethod(servicesProvider: ServiceProvider = ServiceProvider)
   def this() = this(ServiceProvider)
 
   override protected def servicesHandler(state: CreatePaymentMethodState, requestInfo: RequestInfo, context: Context, services: Services) = {
-    logger.debug(s"CreatePaymentMethod state: $state")
+    val someTestVariable = "Happy Tuesday, pal!"
+    SafeLogger.error(scrub"IGNORE ME. Leigh-Anne is learning how to scrub logs sent to sentry. Look at this variable I've logged: $someTestVariable")
```
lmath: I forgot to add exclude the sentry logs in the other appender a la https://github.com/guardian/support-frontend/pull/505 so this log is temporarily here because I want to see that working but otherwise PR comments are very welcome!

common/src/main/scala/com/gu/monitoring/SafeLogger.scala
rupertbates: I think it's fine to leave it here. There are a couple of shared libraries eg. https://github.com/guardian/support-config & https://github.com/guardian/support-models but I don't think it particularly fits in either of those and it is a fairly small piece of code

PR #119, opened 2018-05-24
Fresh Sentry project

lmath:
## Why are you doing this?
We scrubbed PII here: https://github.com/guardian/support-workers/pull/118

So we will point to a new sentry project, and delete the old to start anew. 

[**Trello Card**](https://trello.com/c/LiojyKYR/1555-delete-all-projects-which-contain-pii-from-sentry-and-replace-with-a-clean-project)

## Changes

* Pointing to the DSN of the new project

TODO:
- [x] Test in CODE that error logs go to the new project
- [x] Delete the existing sentry project
- [x] Rename the new sentry project back to support-workers as before


lmath:
PR to upgrade the sentry DSN to follow!

PR #124, opened 2018-07-12
Digipack - Part 1

rupertbates:
## Why are you doing this?
To add Digital Pack as a purchasable product in `support-workers`.

This PR is a refactor to cope with the new Json schema introduced in this PR: https://github.com/guardian/support-models/pull/12. It doesn't fully implement the ability to create a Digital Pack subscription yet, that will come in a separate PR.

I have included a [Circe decoder which can cope with the old schema](https://github.com/guardian/support-workers/pull/124/files#diff-ff423c923c9849219ce186fc2e05b6b3R22) so that we can deploy this without upgrading `support-frontend`.

[**Trello Card**](https://trello.com/c/zHFyq2ly/1523-add-support-for-digi-pack-to-support-workers)

TODO in separate PR:
- [ ] Zuora calls to create a Digital Pack subscription
- [ ] New config for Digi pack
- [ ] Send correct emails for Digi pack


src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -56,51 +57,69 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
-    SubscribeRequest(List(
-      SubscribeItem(
-        account,
-        contactDetails,
-        state.paymentMethod,
-        subscriptionData,
-        SubscribeOptions()
-      )
-    ))
+  private def buildContactDetails(state: CreateZuoraSubscriptionState) = {
```
jacobwinch: 👍for adding these functions

PR #126, opened 2018-07-16
Config fix

rupertbates:
## Why are you doing this?

* I noticed that the private config files currently contain all the Zuora product ids which could be public so I've moved them to the public config to simplify things.
* I also noticed that the `CreateZuoraSubscriptionSpec` tests are not currently testing anything because there is code in ZuoraService to prevent multiple recurring contributions with the same Identity id. To get round this I have mocked the method which checks for a subscription to always return `None`

### DIFF
--- a/src/test/scala/com/gu/support/workers/integration/CreateZuoraSubscriptionSpec.scala
+++ b/src/test/scala/com/gu/support/workers/integration/CreateZuoraSubscriptionSpec.scala
@@ -2,6 +2,8 @@ package com.gu.support.workers.integration
 
 import java.io.ByteArrayOutputStream
 
+import com.gu.config.Configuration.zuoraConfigProvider
+import com.gu.okhttp.RequestRunners.configurableFutureRunner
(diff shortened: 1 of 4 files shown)


src/test/scala/com/gu/support/workers/integration/CreateZuoraSubscriptionSpec.scala
```diff
@@ -2,38 +2,79 @@ package com.gu.support.workers.integration
...
+
+    val sendThankYouEmail = Encoding.in[SendThankYouEmailState](outStream.toInputStream).get
+    sendThankYouEmail._1.accountNumber.length should be > 0
+  }
+
+  val realService = new ZuoraService(zuoraConfigProvider.get(false), configurableFutureRunner(60.seconds))
+
+  val mockService = {
+    val z = mock[ZuoraService]
+    // Need to return None from the Zuora service `getRecurringSubscription`
```
jacobwinch: Woops... good spot 👍

PR #129, opened 2018-07-19
Implement digi pack - part 2

rupertbates:
## Why are you doing this?
This is part 2 of adding the Digital Pack into support workers
* Add a new Zuora call to create digital pack subscriptions - this actually only involves sending the correct `productRatePlanId` for the Digital Pack product instead of the Contribution one. There is new config to support this.
* Add a new email for the Digital Pack, this has involved quite a major refactor of the way that emails are handled. I've introduced separate classes for each email type and removed all the config settings that controlled them previously - these were partly redundant anyway because all the emails use the same SQS queue and now that we have separate classes for each email type the dataExtensionName sits logically in those.
* Use the new state model in this PR: https://github.com/guardian/support-models/pull/38 (**Until this is released the TeamCity build will fail**)

[**Trello Card**](https://trello.com/c/zHFyq2ly/1523-add-support-for-digi-pack-to-support-workers)


src/main/scala/com/gu/emailservices/ContributionEmailFields.scala
```diff
@@ -0,0 +1,41 @@
...
+    directDebitMandateId: Option[String] = None
+) extends EmailFields {
+
+  val paymentFields = paymentMethod match {
+    case Some(dd: DirectDebitPaymentMethod) => List(
+      "account name" -> dd.bankTransferAccountName,
+      "account number" -> mask(dd.bankTransferAccountNumber),
+      "sort code" -> hyphenate(dd.bankCode),
+      "Mandate ID" -> directDebitMandateId.getOrElse(""),
+      "first payment date" -> formatDate(created.plusDays(10)),
```
jacobwinch: Is this correct for a recurring contribution? I thought we tried to take the payment immediately?

src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -0,0 +1,81 @@
...
+    currency: Currency,
+    paymentMethod: PaymentMethod,
+    directDebitMandateId: Option[String] = None
+) extends EmailFields {
+
+  val amount = 11.99F //TODO: Get the real price
+  val subscriptionDetails = s"${currency.glyph}$amount every ${billingPeriod.noun}"
+  val paymentFields = paymentMethod match {
+    case dd: DirectDebitPaymentMethod => List(
+      "Account number" -> mask(dd.bankTransferAccountNumber),
```
jacobwinch: might be worth adding a helper function to build this list (looks like we do a similar thing twice)?

src/main/scala/com/gu/emailservices/EmailService.scala
```diff
@@ -4,84 +4,25 @@ import com.amazonaws.regions.Regions
...
+class EmailService(implicit val executionContext: ExecutionContext) {
 
   private val sqsClient = AmazonSQSAsyncClientBuilder
     .standard
     .withCredentials(CredentialsProvider)
     .withRegion(Regions.EU_WEST_1)
     .build()
 
-  private val queueUrl = sqsClient.getQueueUrl(config.queueName).getQueueUrl
+  private val queueName = "contributions-thanks"
```
jacobwinch: So Digipack emails will also use the contributions-thanks queue now?

src/main/scala/com/gu/emailservices/ContributionEmailFields.scala
rupertbates: It is right for DD, I can't remember exactly but I think it's to give time for the DD to be set up?

src/test/scala/com/gu/support/workers/integration/SendThankYouEmailSpec.scala
```diff
@@ -27,33 +28,51 @@ class SendThankYouEmailSpec extends LambdaSpec {
...
       Currency.GBP,
-      "UK", "", "monthly-contribution", Some(dd), Some(mandateId)
+      "UK", "", Some(dd), Some(mandateId)
     )
-    val service = new EmailService(Configuration.emailServicesConfig.thankYou, executionContext)
+    val service = new EmailService
+    service.send(ef)
+  }
+
+  it should "send a digital pack email" in {
```
jacobwinch: You might want to ignore this test too?

src/main/scala/com/gu/emailservices/ContributionEmailFields.scala
jacobwinch: I guess it makes sense that it would take a number of days for GoCardless to set up the mandate etc. Thanks for clarifying 👍

src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
jacobwinch: Ah yes sorry, I didn't spot the slight distinctions

src/main/scala/com/gu/emailservices/EmailService.scala
rupertbates: Yes as far as I'm aware there's no advantage to having a separate queue for individual email types

PR #130, opened 2018-07-23
send alerts to simple-dev only

michaelwmcnamara:
Currently a state machine failure is broadcast to all of reader revenue dev.
This is overkill as 2/3's of recipients are not going to take any action on the alert.
Moreover this increases the risk of developers "tuning out" AWS alerts as a result of them being frequent and not relevant to their immediate area of responsibility.
The plan going forward is to send alerts that are not time critical to the specific team of responsibility, reserving the reader.revenue.dev email address for time critical "everything is on fire!" style alerts - where we want action to be taken by the first person to see the alert.

### DIFF
--- a/cloud-formation/src/templates/cfn-template.yaml
+++ b/cloud-formation/src/templates/cfn-template.yaml
@@ -125,7 +125,7 @@ Resources:
     Condition: CreateProdResources
     Properties:
       AlarmActions:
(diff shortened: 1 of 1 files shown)


jacobwinch:
🚢

rupertbates:
@michaelwmcnamara what is going on with this PR, is it still relevant or can we close it?

PR #133, opened 2018-08-02
Fix polling functionality for payment failures

jacobwinch:
## Why are you doing this?

As a follow up to:
https://github.com/guardian/support-workers/pull/131

This PR removes the ContributionCompleted lambda (and associated code). We also add a new state so that support-frontend can recognise failed payments again.

Since I had to change the logic of the step function, I've also removed WaitForBrowserPoll (I discussed it with @davidfurey and we agreed that it's no longer needed).

## Trello Cards
[Fix Polling](https://trello.com/c/KTqKRJ44/1826-fix-polling-functionality-to-improve-checkout-experience)
[Remove WaitForBrowserPoll](https://trello.com/c/faO9GSte/1825-remove-unnecessary-waitforbrowserpoll-step-from-support-workers)

## Changes

* Remove all code related to ContributionCompletedLambda
* Remove ContributionCompletedLambda from riff-raff.yaml
* Temporarily stop returning a state from the FailureHandler (I will re-introduce a new state which allows us to pass through specific payment failure messages in a separate PR).
* Fix the tests
* Terminate successful executions immediately after running the Parallel block 
* Terminate checkout failures immediately after making the succeed/fail choice

## Checkout with successful payment
image

## Checkout with payment failure
image

## Checkout with unknown failure
image


src/main/scala/com/gu/support/workers/lambdas/FailureHandler.scala
```diff
@@ -39,34 +39,24 @@ class FailureHandler(emailService: EmailService)
...
-        "There was an error processing your payment. Please\u00a0try\u00a0again\u00a0later.")
+      case _ => exitHandler(state, requestInfo.copy(failed = true))
     }
   }
 
-  private def returnState(
+  private def exitHandler(
     state: FailureHandlerState,
-    requestInfo: RequestInfo,
-    message: String = "There was an error processing your payment. Please\u00a0try\u00a0again."
```
jacobwinch: We always just display the default failure message client-side, so we're not really losing anything by removing this state at the moment...

I'd like to re-introduce a CheckoutFailure state in a separate PR, which can be used to pass more specific errors back to the client.

src/test/scala/com/gu/support/workers/EndToEndSpec.scala
```diff
@@ -27,7 +27,10 @@ class EndToEndSpec extends LambdaSpec {
       .chain(new CreatePaymentMethod())
       .chain(new CreateSalesforceContact())
       .chain(new CreateZuoraSubscription())
-      .parallel(new ContributionCompleted, new SendThankYouEmail())
+      .parallel(
+        new SendThankYouEmail(),
+        new SendAcquisitionEvent(MockOphanHelper.mockServices) //We have to mock Ophan - there doesn't seem to be a test mode
```
jacobwinch: I noticed that this wasn't included in the test, but thought it would be useful to continue testing parallel steps, even if one is just calling a mock?

PR #134, opened 2018-08-07
Send identity id through in acquisition event

jranks123:
## Why are you doing this?

Previously, when running the `acquisitions` job in the excavator, we relied on joining on the `pageview` table to get the `identity id` for recurring contributions. Seeing as it is available here, we should send it through, lessening our dependence on the `pageview` table.

@guardian/contributions


jranks123:
@joelochlann yes we do

PR #135, opened 2018-08-09
On failure, send a digipack specific email for digipack

lmath:
## Why are you doing this?

https://github.com/guardian/support-workers/pull/124
https://github.com/guardian/support-workers/pull/129

⬆️ Work done so that we can process digipack signups async as we do for recurring contributions. 

However, sometimes we have told the user the signup was successful when it has actually eventually failed. So, we need to send them a _digipack specific_  failure email asking them to try again.

[**Trello Card**](https://trello.com/c/abpHEM88/1798-failed-subscription-sign-up-email)

## Changes
* In https://github.com/guardian/membership-common/pull/581, https://github.com/guardian/membership-workflow/pull/142 we added a digipack specific failure email. So here we add FailedDigitalPackEmailFields to send that email when appropriate
* Match on product in the failure handler and send the appropriate email.

## Email screenshot
image


lmath:
I struggled to find the best way to convince myself that the failure handler was indeed matching correctly on product and sending the right email. In the end, I added a [test case](https://github.com/guardian/support-workers/pull/135/files#diff-6a87ba509de9f9f70a3bcc0b92f3b6d9R77 ) with stripe payment errors for a digital pack so that I could unsuppress the logging and see that the failure handler still picked FailedContributionEmailFields for a failed contribution and FailedDigitalPackEmailFields for a failed digital pack signup. Having to manually check the logging is clearly not ideal so any suggestions very much welcome!

rupertbates:
@lmath To test that match you could add a test which passes a mock `EmailService` into the `FailureHandler` and verify that `emailService.send` is called with the correct type of object

jacobwinch:
Looks good 👍

lmath:
Thanks @rupertbates ! I think I got confused about how the FailureHandler gets an EmailService even though it just... instantiates one. Oops. I've put in a mock so I can check sendEmail gets called with FailedDigitalPackEmailFields.

PR #136, opened 2018-08-14
Provide clients with a reason for a failed checkout

jacobwinch:
## Why are you doing this?

Exits the FailureHandler with a reason from: https://github.com/guardian/support-models/pull/40

Note that the build for this PR will fail until support-models has been released.

[**Trello Card**](https://trello.com/c/Q2E9wl9R/1836-return-specific-errors-based-on-payment-failure-reason)

## Changes

* Add JSON encoders/decoders for new case classes
* Exit the handler with a sensible error


src/main/scala/com/gu/support/workers/encoding/StateCodecs.scala
```diff
@@ -13,8 +14,15 @@ object StateCodecs {
 
   implicit val encodeStatus: Encoder[Status] = Encoder.encodeString.contramap[Status](_.asString)
 
-  implicit val decodeStatus: Decoder[Status] =
-    Decoder.decodeString.emap { identifier => Status.fromString(identifier).toRight(s"Unrecognised status '$identifier'") }
+  implicit val encodeFailureReason: Encoder[CheckoutFailureReason] = Encoder.encodeString.contramap[CheckoutFailureReason](_.asString)
```
rupertbates: I'd move `decodeStatus` after `encodeStatus` and `decodeFailureReason` after `encodeFailureReason`

src/main/scala/com/gu/support/workers/lambdas/FailureHandler.scala
```diff
@@ -71,4 +79,15 @@ class FailureHandler(emailService: EmailService)
   } yield result
 
   private def tryToDecode[T](errorJson: ErrorJson)(implicit decoder: Decoder[T]): Option[T] = decode[T](errorJson.errorMessage).toOption
+
+  def toCheckoutFailureReason(zuoraError: ZuoraError): CheckoutFailureReason = {
+    // Just get Stripe's decline code (example message from Zuora: "Transaction declined.do_not_honor - Your card was declined.")
+    val trimmedError = zuoraError.Message.stripPrefix("Transaction declined.").split(" ")(0)
```
rupertbates: I take it this is the only way to get this code? It seems pretty nasty.

src/test/scala/com/gu/support/workers/lambdas/FailureHandlerSpec.scala
```diff
@@ -120,4 +119,11 @@ class FailureHandlerSpec extends LambdaSpec {
       case _ => fail()
     }
   }
+
+  it should "convert a transaction declined error from Zuora to an appropriate CheckoutFailureReason" in {
+    val failureHandler = new FailureHandler()
+    val reason = failureHandler.toCheckoutFailureReason(ZuoraError("TRANSACTION_FAILED", "Transaction declined.do_not_honor - Your card was declined."))
+    reason should be(PaymentMethodUnacceptable)
+  }
+
```
rupertbates: It would be good to check the reason returned in all the other test cases in this spec.

src/main/scala/com/gu/support/workers/lambdas/FailureHandler.scala
jacobwinch: Yes, unfortunately I think this is the only way to get it without making additional API calls (and that would become quite complicated)

jacobwinch: It's a shame that Zuora don't just return the decline_code directly from Stripe...

src/test/scala/com/gu/support/workers/lambdas/FailureHandlerSpec.scala
```diff
@@ -63,9 +74,13 @@ class FailureHandlerSpec extends LambdaSpec {
     val outState = decode[JsonWrapper](Source.fromInputStream(outStream.toInputStream).mkString)
```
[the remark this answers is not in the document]
jacobwinch: Ah yes, thanks - this is much neater 👍

PR #138, opened 2018-08-30
set product in email fields

tsop14:
## Why are you doing this?
So exact target can send thank you emails tailored to frequency i.e. annual and monthly contributions. 📨

[**Trello Card**](https://trello.com/c/6qGb9bHv/751-validate-monthly-and-annual-email-confirmations)

## Changes

* Adds product field to ContributionEmailFields class
* Sets value using `state.product.billingPeriod` which will result in: `Monthly-contribution` or `Annual-contribution`being passed through. 
* Updates test objects to include product field. 

### TODO

- [ ] Use this [test](https://github.com/guardian/support-workers/blob/master/src/test/scala/com/gu/support/workers/integration/SendThankYouEmailSpec.scala#L37) once exact target has been updated to send different templates according to the product value, to ensure correct emails are sent.

### DIFF
--- a/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
+++ b/src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
@@ -49,6 +49,7 @@ class SendThankYouEmail(thankYouEmailService: EmailService, servicesProvider: Se
           currency = c.currency,
           edition = state.user.country.alpha2,
           name = state.user.firstName,
+          product = s"${state.product.billingPeriod}-contribution",
           paymentMethod = Some(state.paymentMethod),
           directDebitMandateId = directDebitMandateId
         )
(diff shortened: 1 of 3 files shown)


src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
```diff
@@ -49,6 +49,7 @@ class SendThankYouEmail(thankYouEmailService: EmailService, servicesProvider: Se
           currency = c.currency,
           edition = state.user.country.alpha2,
           name = state.user.firstName,
+          product = s"${state.product.billingPeriod}-contribution",
```
rupertbates: I would pass the billing period into the `ContributionEmailFields` class and generate this string there as there is less opportunity for a calling function to introduce variations that way. 

Also probably lowercase the billing period in the generated string to match with the previous hard coded casing.

tsop14: 👍

PR #139, opened 2018-09-19
Add sfContactId to Email payload

QuarpT:
## Changes

Add sfContactId and Identity userId to the email payload.  The messages are consumed by membership-workflow which triggers the emails. In Braze emails we upsert users using the sfContactId as the primary key. membership-workflow can resolve the sfContactId from the identity user id

## Why are you doing this?

We are migrating from our CMT providers, from ET to Braze

### DIFF
--- a/src/main/scala/com/gu/emailservices/EmailFields.scala
+++ b/src/main/scala/com/gu/emailservices/EmailFields.scala
@@ -6,11 +6,13 @@ import io.circe.generic.auto._
 import io.circe.syntax._
 import com.gu.salesforce.Salesforce.SfContactId
 
+// scalastyle:off
 case class EmailPayloadContactAttributes(SubscriberAttributes: Map[String, String])
 case class EmailPayloadTo(Address: String, SubscriberKey: String, ContactAttributes: EmailPayloadContactAttributes)
 case class EmailPayload(To: EmailPayloadTo, DataExtensionName: String, SfContactId: Option[String], IdentityUserId: Option[String]) {
   lazy val jsonString: String = this.asJson.toString
 }
+// scalastyle:on
 
 case class IdentityUserId(id: String)
 
--- a/src/main/scala/com/gu/emailservices/EmailFields.scala
+++ b/src/main/scala/com/gu/emailservices/EmailFields.scala
@@ -4,6 +4,7 @@ import org.joda.time.DateTime
 import org.joda.time.format.DateTimeFormat
 import io.circe.generic.auto._
 import io.circe.syntax._
+import com.gu.salesforce.Salesforce.SfContactId
 
 case class EmailPayloadContactAttributes(SubscriberAttributes: Map[String, String])
 case class EmailPayloadTo(Address: String, SubscriberKey: String, ContactAttributes: EmailPayloadContactAttributes)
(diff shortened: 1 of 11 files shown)


src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
```diff
@@ -50,6 +50,8 @@ class SendThankYouEmail(thankYouEmailService: EmailService, servicesProvider: Se
           edition = state.user.country.alpha2,
           name = state.user.firstName,
           billingPeriod = state.product.billingPeriod,
+          sfContactId = Some(state.salesForceContact.Id),
+          identityId = None,
```
rupertbates: So if I understand correctly you only need *either* the sfContactId or the identity id but not both?
In which case it might be nice to use an `Either` rather than two options?

src/main/scala/com/gu/emailservices/EmailFields.scala
```diff
@@ -1,35 +1,39 @@
 package com.gu.emailservices
 
-import io.circe.{Json, Printer}
 import org.joda.time.DateTime
 import org.joda.time.format.DateTimeFormat
+import io.circe.generic.auto._
+import io.circe.syntax._
+
+case class EmailPayloadContactAttributes(SubscriberAttributes: Map[String, String])
```
rupertbates: Do you think these classes are adding anything? Personally I prefer being able to see the full structure of the final json rather than having to infer it from a class hierarchy. Were you looking for improved compile time verification or something?

johnduffell: doing your own serialisation can look nice, but it can be quite dangerous because now you are responsible for all your own escaping.  a hopefully well tested library to do the serialisation to string is safer.  I think 3 of the top 10 including number 1 OWASP top 10 are caused by this kind of thing.

src/main/scala/com/gu/emailservices/EmailFields.scala
```diff
@@ -1,35 +1,41 @@
 package com.gu.emailservices
 
-import io.circe.{Json, Printer}
 import org.joda.time.DateTime
 import org.joda.time.format.DateTimeFormat
+import io.circe.generic.auto._
+import io.circe.syntax._
+import com.gu.salesforce.Salesforce.SfContactId
+
+case class EmailPayloadContactAttributes(SubscriberAttributes: Map[String, String])
```
johnduffell: the comments on this line seem to have been collapsed for some reason https://github.com/guardian/support-workers/pull/139#pullrequestreview-156844346

src/main/scala/com/gu/emailservices/EmailFields.scala
QuarpT: The advantage is compile time type safety. You're less likely to make mistakes by relying on the circe serializer typeclasses compared to building the json string dynamically using custom logic (I don't want to write code for converting optional fields to json!).

I don't see the difficulties reading a simple case class hierarchy. The json conversion is trivial.

src/main/scala/com/gu/emailservices/EmailFields.scala
QuarpT: I've replied 😄

src/main/scala/com/gu/support/workers/lambdas/SendThankYouEmail.scala
QuarpT: I made the changes in f05ab53cd3c84756791b406aece73f8fa4049368

PR #140, opened 2018-10-02
configurable thank you email queue name

QuarpT:
## Why are you doing this?

We need to be able to configure CODE to use the CODE queues. This is necessary for testing Braze emails sent by membership-workflow in CODE.

## Changes

* configurable thank you email queue name


src/main/scala/com/gu/config/Configuration.scala
```diff
@@ -29,4 +29,6 @@ object Configuration {
   val payPalConfigProvider = new PayPalConfigProvider(config, stage)
   val salesforceConfigProvider = new SalesforceConfigProvider(config, stage)
   val zuoraConfigProvider = new ZuoraConfigProvider(config, stage)
+
+  val contributionThanksQueueName = config.getString("email.thankYou.queueName")
```
paulbrown1982: Should you add a default value to https://github.com/guardian/support-workers/blob/master/src/main/resources/application.conf ?

PR #141, opened 2018-10-04
send payment type to email send sqs queue

QuarpT:
### Why are you doing this?

The service email templates will render content depending on payment type

### What are the changes?

Add payment method name to SubscriberAttributes sent to Exact Target and Braze. These are used by the email templates.

### DIFF
--- a/src/main/scala/com/gu/emailservices/ContributionEmailFields.scala
+++ b/src/main/scala/com/gu/emailservices/ContributionEmailFields.scala
@@ -1,7 +1,7 @@
 package com.gu.emailservices
 
 import com.gu.i18n.Currency
-import com.gu.support.workers.model.{BillingPeriod, DirectDebitPaymentMethod, PaymentMethod}
+import com.gu.support.workers.model._
 import org.joda.time.DateTime
 import com.gu.salesforce.Salesforce.SfContactId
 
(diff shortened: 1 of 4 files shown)


src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
```diff
@@ -57,7 +57,7 @@ case class DigitalPackEmailFields(
       "MandateID" -> directDebitMandateId.getOrElse("")
     )
     case _: CreditCardReferenceTransaction => List("Default payment method" -> "Credit/Debit Card")
-    case _ => Seq("Default payment method" -> "PayPal")
+    case _: PayPalReferenceTransaction => Seq("Default payment method" -> "PayPal")
```
QuarpT: payment method is already being sent for the DigitalPackEmailFields. The values follow a different format `"Credit/Debit Card"`.

src/main/scala/com/gu/emailservices/ContributionEmailFields.scala
```diff
@@ -14,20 +14,21 @@ case class ContributionEmailFields(
     name: String,
     billingPeriod: BillingPeriod,
     sfContactId: SfContactId,
-    paymentMethod: Option[PaymentMethod] = None,
+    paymentMethod: PaymentMethod,
```
johnduffell: good spot!

src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
johnduffell: ok

I just look at the rest of the file and discovered it's full of commented out JSON and TODOs 😨 however I don't think it's been further invalidated by this change

johnduffell:
looks good to me 👍

src/main/scala/com/gu/emailservices/DigitalPackEmailFields.scala
johnduffell: got it! thanks for the explanation @rupertbates

PR #143, opened 2018-10-10
allow creation of a second recurring contribution after 10 minutes

johnduffell:
## Why are you doing this?

At present there is logic in the payment flow to stop people adding a second recurring contribution.  There is also logic in the step functions to stop us adding a second zuora contribution record.  This is necessary because sometimes the zuora step isn't confirmed to have succeeded but it has actually happened in the background.  Then the retry would add an incorrect duplication.

The new payment flow doesn't require login, just the email, so we need to not leak any information to the person entering the data.

This PR relaxes the corresponding lambda check to only search for subscriptions created in the last 10 minutes.

This means that we should still catch all the retries, but removing most of the opportunity for data leakage.

[**Trello Card**](https://trello.com/c/eJnOL8se/71-make-step-functions-allow-multiple-recurring-contributions-for-one-identity-account)

## Changes

@jacobwinch @rupertbates @joelochlann @jranks123 @tsop14 let me know what you think!

### DIFF
--- a/src/main/scala/com/gu/support/workers/lambdas/GetRecurringSubscription.scala
+++ b/src/main/scala/com/gu/support/workers/lambdas/GetRecurringSubscription.scala
@@ -1,0 +1,46 @@
+package com.gu.support.workers.lambdas
+
+import java.time.LocalDateTime
+
+import cats.implicits._
+import com.gu.support.workers.model.BillingPeriod
+import com.gu.zuora.GetAccountForIdentity.DomainAccount
+import com.gu.zuora.ZuoraConfig.RatePlanId
+import com.gu.zuora.ZuoraService
+import com.gu.zuora.model.response.{RatePlan, Subscription}
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+object GetRecurringSubscription {
+
+  def apply(zuoraService: ZuoraService, now: LocalDateTime, identityId: String, billingPeriod: BillingPeriod): Future[Option[Subscription]] = {
+
+    val productRatePlanId = Option(zuoraService.config).map(_.contributionConfig(billingPeriod).productRatePlanId)
+
+    def hasContributorPlan(sub: Subscription): Boolean =
+      productRatePlanId match {
+        case Some(productRatePlanId) => GetRecurringSubscription.hasContributorPlan(productRatePlanId, sub.ratePlans)
+        case None => false
+      }
+
+    def isRecent(domainAccount: DomainAccount): Boolean =
+      GetRecurringSubscription.isRecent(now, domainAccount.createdDate)
+
+    for {
+      accountIds <- zuoraService.getAccountFields(identityId)
+      recentAccountIds = accountIds.filter(isRecent).map(_.accountNumber)
+      subscriptions <- recentAccountIds.map(zuoraService.getSubscriptions).combineAll
+      maybeRecentCont = subscriptions.find(sub => hasContributorPlan(sub) && sub.isActive)
+    } yield maybeRecentCont
+  }
+
+  def isRecent(now: LocalDateTime, accountCreation: LocalDateTime): Boolean = {
+    val oldestRecentTime = now.minusMinutes(10L)
+    accountCreation.isAfter(oldestRecentTime)
+  }
+
+  def hasContributorPlan(ratePlanId: RatePlanId, ratePlans: List[RatePlan]): Boolean =
+    ratePlans.exists(_.productRatePlanId == ratePlanId)
+
+}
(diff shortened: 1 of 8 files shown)


src/main/scala/com/gu/support/workers/lambdas/GetRecurringSubscription.scala
```diff
@@ -0,0 +1,46 @@
...
+import com.gu.zuora.ZuoraConfig.RatePlanId
+import com.gu.zuora.ZuoraService
+import com.gu.zuora.model.response.{RatePlan, Subscription}
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+object GetRecurringSubscription {
+
+  def apply(zuoraService: ZuoraService, now: LocalDateTime, identityId: String, billingPeriod: BillingPeriod): Future[Option[Subscription]] = {
```
johnduffell: I've just split this from the zuora service as it was just packing that class out with unrelated code and making it harder to test.  It shouldn't really depend on everything inside zuoraService object, but that kind of decoupling is not a task to be taken lightly in this PR.

src/main/scala/com/gu/zuora/ZuoraConfig.scala
```diff
@@ -9,7 +9,7 @@ object ZuoraConfig {
   type RatePlanId = String
 }
 
-case class ZuoraContributionConfig(productRatePlanId: RatePlanId, productRatePlanChargeId: RatePlanId)
+case class ZuoraContributionConfig(productRatePlanId: RatePlanId, productRatePlanChargeId: RatePlanId /*really?*/ )
```
johnduffell: I don't think this is a RatePlanId (although it can be represented as a String).  Wasn't sure whether to refactor it in this pr though?

src/main/scala/com/gu/zuora/ZuoraService.scala
```diff
@@ -70,3 +66,19 @@ class ZuoraService(config: ZuoraConfig, client: FutureHttpClient, baseUrl: Optio
...
+object GetAccountForIdentity {
+
+  case class DomainAccount(accountNumber: String, createdDate: LocalDateTime)
+
+  object DomainAccount {
+
+    def fromAccountRecord(accountRecord: AccountRecord): DomainAccount =
+      DomainAccount(
+        accountRecord.AccountNumber,
+        ZonedDateTime.parse(accountRecord.CreatedDate).withZoneSameInstant(ZoneId.systemDefault()).toLocalDateTime
```
johnduffell: zuora returns a zoned date time with UTC+0.  I want to parse it properly, but I wasn't sure whether to convert that to local time, or to use ZonedDateTime.now to do the comparison.  I started to feel like the latter would be better

src/test/scala/com/gu/zuora/ZuoraSpec.scala
```diff
@@ -46,25 +49,31 @@ class ZuoraSpec extends AsyncFlatSpec with Matchers {
...
 
-  it should "be able to find a monthly recurring subscription" in {
-    uatService.getRecurringSubscription("30001758", Monthly).map {
-      response =>
-        response.isDefined should be(true)
-        response.get.ratePlans.head.productName should be("Contributor")
+  it should "be able to find a monthly recurring subscription just after creation" in {
+    val justAfterCreation = LocalDateTime.of(2017, 12, 7, 15, 47, 22)
+    GetRecurringSubscription(uatService, justAfterCreation, "30001758", Monthly).map {
+      _.flatMap(_.ratePlans.headOption.map(_.productName)) should be(Some("Contributor"))
```
johnduffell: this was just because I was getting "false should be true" type messages from the tests which was a bit unsatisfying

jacobwinch:
> This is necessary because sometimes the zuora step isn't confirmed to have succeeded but it has actually happened in the background. Then the retry would add an incorrect duplication.

The stricter version of the check which is being removed here has some additional benefits:

1. It allows us to have more certainty that users only have one recurring contribution (which makes life simpler when thinking about account management etc.)
2. It provides a safety mechanism against charging people repeatedly (if we introduce a bug which causes this lambda to be invoked multiple times)

I appreciate there's a data privacy concern here, so 👍if we have to lose 1 in order to address that. 

I would prefer to retain 2 if possible, so perhaps we can increase this window to 24 hours (to match the max execution time https://github.com/guardian/support-workers/blob/master/cloud-formation/src/templates/state-machine.yaml#L3).

johnduffell:
Ok after discussion with Jonny, we need to create a unique token on every signup page and store that in zuora, otherwise there's no way round it.  I have added the details to the doc https://docs.google.com/document/d/14wYvCij4pYZ-0YIY46lO4DZhL4ZfHD2max6WAyLnkpQ/edit#bookmark=id.ukf598745msh

PR #147, opened 2018-10-19
use the session id to decide which zuora accounts to deduplicate against

johnduffell:
## Why are you doing this?

This PR is part of the series designed to deduplicate zuora subscriptions in case of browser side or step function repetitions.  I have already changed the models https://github.com/guardian/support-models/pull/43 and support frontend https://github.com/guardian/support-frontend/pull/1049

[**Trello Card**](https://trello.com/c/qaNYnBkp/87-npf-make-step-functions-check-form-visit-token-as-well-as-identity-id-for-preventing-duplicate-recurring-cont-from-the-same-page)

I have tested this in CODE but need to redo with the final approved code before shipping.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -43,7 +43,7 @@ lazy val root = (project in file("."))
       "ch.qos.logback" % "logback-classic" % "1.2.3",
       "io.symphonia" % "lambda-logging" % "1.0.1",
       "com.gu" %% "support-internationalisation" % "0.9",
-      "com.gu" %% "support-models" % "0.34",
+      "com.gu" %% "support-models" % "0.35-SNAPSHOT",
       "com.gu" %% "support-config" % "0.16",
       "com.squareup.okhttp3" % "okhttp" % okhttpVersion,
       "com.netaporter" %% "scala-uri" % "0.4.16",
--- a/build.sbt
+++ b/build.sbt
@@ -43,7 +43,7 @@ lazy val root = (project in file("."))
       "ch.qos.logback" % "logback-classic" % "1.2.3",
       "io.symphonia" % "lambda-logging" % "1.0.1",
       "com.gu" %% "support-internationalisation" % "0.9",
-      "com.gu" %% "support-models" % "0.35-SNAPSHOT",
+      "com.gu" %% "support-models" % "0.35",
       "com.gu" %% "support-config" % "0.16",
       "com.squareup.okhttp3" % "okhttp" % okhttpVersion,
       "com.netaporter" %% "scala-uri" % "0.4.16",
--- a/build.sbt
+++ b/build.sbt
@@ -3,7 +3,7 @@ import com.typesafe.sbt.SbtScalariform.ScalariformKeys
 import sbt.Keys.{libraryDependencies, resolvers}
 import scalariform.formatter.preferences.SpacesAroundMultiImports
 
-scalaVersion := "2.12.4"
+scalaVersion := "2.12.7"
 organization := "com.gu"
 version := "0.1-SNAPSHOT"
 scalacOptions ++= Seq("-deprecation", "-feature", "-unchecked", "-target:jvm-1.8", "-Xfatal-warnings")
@@ -43,7 +43,7 @@ lazy val root = (project in file("."))
       "ch.qos.logback" % "logback-classic" % "1.2.3",
       "io.symphonia" % "lambda-logging" % "1.0.1",
       "com.gu" %% "support-internationalisation" % "0.9",
-      "com.gu" %% "support-models" % "0.32",
+      "com.gu" %% "support-models" % "0.34",
       "com.gu" %% "support-config" % "0.16",
       "com.squareup.okhttp3" % "okhttp" % okhttpVersion,
       "com.netaporter" %% "scala-uri" % "0.4.16",
--- a/src/main/scala/com/gu/support/workers/GetRecurringSubscription.scala
+++ b/src/main/scala/com/gu/support/workers/GetRecurringSubscription.scala
@@ -22,13 +22,9 @@ object GetRecurringSubscription {
     billingPeriod: BillingPeriod
   )(implicit ec: ExecutionContext): Future[Option[DomainSubscription]] = {
 
-    val productRatePlanId = Option(zuoraService.config).map(_.contributionConfig(billingPeriod).productRatePlanId)
+    val productRatePlanId: RatePlanId = zuoraService.config.contributionConfig(billingPeriod).productRatePlanId
 
-    def hasContributorPlan(sub: DomainSubscription): Boolean =
-      productRatePlanId match {
-        case Some(productRatePlanId) => GetRecurringSubscription.hasContributorPlan(productRatePlanId, sub.ratePlans)
-        case None => false
-      }
+    val hasContributorPlan: List[RatePlan] => Boolean = GetRecurringSubscription.hasContributorPlan(productRatePlanId)
 
     def isInScope(domainAccount: DomainAccount): Boolean = {
       val inScope = GetRecurringSubscription.isInAccessScope(accessScope, domainAccount.maybeCreatedSessionId)
@@ -40,8 +36,8 @@ object GetRecurringSubscription {
       accountIds <- zuoraService.getAccountFields(identityId)
       inScopeAccountIds = accountIds.filter(isInScope).map(_.accountNumber)
       subscriptions <- inScopeAccountIds.map(zuoraService.getSubscriptions).combineAll
-      maybeRecentCont = subscriptions.find(sub => hasContributorPlan(sub) && sub.isActive.value)
-    } yield maybeRecentCont
+      maybeRecentContributor = subscriptions.find(sub => hasContributorPlan(sub.ratePlans) && sub.isActive.value)
+    } yield maybeRecentContributor
   }
 
   def isInAccessScope(accessScope: AccessScope, maybeCreatedSessionId: Option[SessionId]): Boolean = {
@@ -52,7 +48,7 @@ object GetRecurringSubscription {
     }
   }
 
-  def hasContributorPlan(ratePlanId: RatePlanId, ratePlans: List[RatePlan]): Boolean =
+  def hasContributorPlan(ratePlanId: RatePlanId)(ratePlans: List[RatePlan]): Boolean =
     ratePlans.exists(_.productRatePlanId == ratePlanId)
 
 }
--- a/src/main/scala/com/gu/support/workers/GetRecurringSubscription.scala
+++ b/src/main/scala/com/gu/support/workers/GetRecurringSubscription.scala
@@ -1,0 +1,54 @@
+package com.gu.support.workers
+
+import cats.implicits._
+import com.gu.support.workers.lambdas.IdentityId
+import com.gu.support.workers.model.AccessScope.{AccessScopeBySessionId, AccessScopeNoRestriction, SessionId}
+import com.gu.support.workers.model.{AccessScope, BillingPeriod}
+import com.gu.zuora.GetAccountForIdentity.DomainAccount
+import com.gu.zuora.GetSubscription.DomainSubscription
+import com.gu.zuora.ZuoraConfig.RatePlanId
+import com.gu.zuora.ZuoraService
+import com.gu.zuora.model.response.RatePlan
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object GetRecurringSubscription {
+
+  def apply(
+    zuoraService: ZuoraService,
+    accessScope: AccessScope,
+    identityId: IdentityId,
+    billingPeriod: BillingPeriod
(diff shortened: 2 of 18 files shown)


src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -24,45 +28,53 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
     requestInfo: RequestInfo,
     context: Context,
     services: Services
-  ): FutureHandlerResult =
-    services.zuoraService.getRecurringSubscription(state.user.id, state.product.billingPeriod).flatMap {
-      case Some(sub) => skipSubscribe(state, requestInfo, sub)
+  ): FutureHandlerResult = for {
+    identityId <- Future.fromTry(IdentityId(state.user.id))
```
johnduffell: it would be better if all the converting from the Wire model (Strings etc) to the domain model (IdentityId etc) would happen up front to keep things simple at this level.  But I wanted to just do the simplest change so I am just flatmapping in the one field.

src/main/scala/com/gu/zuora/ZuoraService.scala
```diff
@@ -27,24 +30,22 @@ class ZuoraService(config: ZuoraConfig, client: FutureHttpClient, baseUrl: Optio
   def getAccount(accountNumber: String): Future[GetAccountResponse] =
     get[GetAccountResponse](s"accounts/$accountNumber", authHeaders)
 
-  def getAccountIds(identityId: String): Future[List[String]] = {
-    val queryData = QueryData(s"select AccountNumber from account where IdentityId__c = '${identityId.toLong}'")
-    postJson[AccountQueryResponse](s"action/query", queryData.asJson, authHeaders).map(_.records.map(_.AccountNumber))
+  def getAccountFields(identityId: IdentityId): Future[List[DomainAccount]] = {
+    // WARNING constructing queries from strings is inherently dangerous.  Be very careful.
+    val queryData = QueryData(s"select AccountNumber, CreatedSessionId__c from account where IdentityId__c = '${identityId.value}'")
```
johnduffell: the identity id is verified when it is put into the type.  I could re-add the toLong here for safety though?  Really for proper abstraction we need to use the SafeQueryBuilder class which basically escapes everything automatically https://github.com/guardian/zuora-auto-cancel/blob/master/lib/zuora/src/main/scala/com/gu/util/zuora/SafeQueryBuilder.scala

src/main/scala/com/gu/zuora/ZuoraService.scala
```diff
@@ -70,3 +71,36 @@ class ZuoraService(config: ZuoraConfig, client: FutureHttpClient, baseUrl: Optio
...
+
+  case class ZuoraIsActive(value: Boolean) extends AnyVal
+
+  case class DomainSubscription(accountNumber: ZuoraAccountNumber, isActive: ZuoraIsActive, ratePlans: List[RatePlan])
+  object DomainSubscription {
+    def fromWireSubscription(subscription: Subscription): DomainSubscription =
+      DomainSubscription(
+        ZuoraAccountNumber(subscription.accountNumber),
+        ZuoraIsActive(subscription.status == "Active"),
+        subscription.ratePlans //.map(DomainRatePlan.fromWire)
```
johnduffell: no need to make RatePlan into a domain model at this stage because I didn't really change that area or notice any specific benefits from adding the type safety now.  Happy to reconsider otherwise I'll delete the comment.

src/main/scala/com/gu/zuora/model/response/SubscriptionsResponse.scala
```diff
@@ -19,12 +19,7 @@ object RatePlan {
 
 case class SubscriptionsResponse(subscriptions: List[Subscription])
 
-case class Subscription(accountNumber: String, status: String, ratePlans: List[RatePlan]) {
-  def isActive: Boolean = status == "Active"
-  def hasContributorPlan(config: ZuoraConfig, billingPeriod: BillingPeriod): Boolean = {
-    ratePlans.exists(_.productRatePlanId == config.contributionConfig(billingPeriod).productRatePlanId)
-  }
-}
+case class Subscription(accountNumber: String, status: String, ratePlans: List[RatePlan])
```
johnduffell: I have made a DomainSubscription that does all these defs in the conversion, so no need the internal methods now.

src/main/scala/com/gu/zuora/ZuoraService.scala
```diff
@@ -70,3 +71,36 @@ class ZuoraService(config: ZuoraConfig, client: FutureHttpClient, baseUrl: Optio
     decode[List[ZuoraErrorResponse]](responseBody).map(_.head)
 
 }
+
+object GetAccountForIdentity {
+
+  case class ZuoraAccountNumber(value: String)
```
johnduffell: this should extend AnyVal, but mockito seems to blow up with an NPE if I do that on https://github.com/guardian/support-workers/pull/147/files#diff-94103699f37bee6329145bc9bae91069R73

src/main/scala/com/gu/support/workers/GetRecurringSubscription.scala
```diff
@@ -0,0 +1,54 @@
...
+import com.gu.support.workers.model.{AccessScope, BillingPeriod}
+import com.gu.zuora.GetAccountForIdentity.DomainAccount
+import com.gu.zuora.GetSubscription.DomainSubscription
+import com.gu.zuora.ZuoraConfig.RatePlanId
+import com.gu.zuora.ZuoraService
+import com.gu.zuora.model.response.RatePlan
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object GetRecurringSubscription {
```
johnduffell: this is basically just the def getRecurringSubscription from zuoraService only extended as necessary, and split into some pure functions as I needed to unit test isInAccessScope

src/main/scala/com/gu/zuora/ZuoraService.scala
```diff
@@ -27,24 +30,22 @@ class ZuoraService(config: ZuoraConfig, client: FutureHttpClient, baseUrl: Optio
...
+  def getAccountFields(identityId: IdentityId): Future[List[DomainAccount]] = {
+    // WARNING constructing queries from strings is inherently dangerous.  Be very careful.
+    val queryData = QueryData(s"select AccountNumber, CreatedSessionId__c from account where IdentityId__c = '${identityId.value}'")
+    postJson[AccountQueryResponse](s"action/query", queryData.asJson, authHeaders).map(_.records.map(DomainAccount.fromWireAccount))
   }
 
-  def getSubscriptions(accountId: String): Future[List[Subscription]] =
-    get[SubscriptionsResponse](s"subscriptions/accounts/$accountId", authHeaders).map(_.subscriptions)
+  def getSubscriptions(accountNumber: ZuoraAccountNumber): Future[List[DomainSubscription]] =
+    get[SubscriptionsResponse](s"subscriptions/accounts/${accountNumber.value}", authHeaders).map { subscriptionsResponse =>
```
johnduffell: it's also dangerous to build urls using string concatenation, but admittedly less, as zuora probably won't trust `/../../` in urls, but if it did, and we POSTed something untrusted at the URL level, it might not be ideal.

src/main/scala/com/gu/support/workers/GetRecurringSubscription.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  def apply(
+    zuoraService: ZuoraService,
+    accessScope: AccessScope,
+    identityId: IdentityId,
+    billingPeriod: BillingPeriod
+  )(implicit ec: ExecutionContext): Future[Option[DomainSubscription]] = {
+
+    val productRatePlanId = Option(zuoraService.config).map(_.contributionConfig(billingPeriod).productRatePlanId)
```
johnduffell: no idea why I did Option @tomrf1 , just pushed a new commit to fix that

src/main/scala/com/gu/support/workers/GetRecurringSubscription.scala
```diff
@@ -0,0 +1,54 @@
...
+
+    def hasContributorPlan(sub: DomainSubscription): Boolean =
+      productRatePlanId match {
+        case Some(productRatePlanId) => GetRecurringSubscription.hasContributorPlan(productRatePlanId, sub.ratePlans)
+        case None => false
+      }
+
+    def isInScope(domainAccount: DomainAccount): Boolean =
```
jranks123: Could we possibly get a comment describing what `isInScope` means? Maybe it's just me but I'm finding it a little hard to folow

johnduffell: I have separated this into a function `IsSubscriptionAccessAllowed` and switched from using Option to an equivalent named trait ExistingContributionSessionId.  I feel like this is much clearer, but have a look and see if it can be improved more.

src/test/scala/com/gu/support/workers/lambdas/CreatePaymentMethodStateDecoderSpec.scala
```diff
@@ -46,31 +47,27 @@ class CreatePaymentMethodStateDecoderSpec extends FlatSpec with Matchers with Mo
...
-    }
+    val maybeState = decode[CreatePaymentMethodState](createPayPalPaymentMethodContributionJson())
+
+    val fieldsToTest = maybeState.map(state =>
+      (state.accessScope, state.product, state.paymentFields))
+    fieldsToTest should be(Right(
+      AccessScopeNoRestriction,
+      Contribution(5, GBP, Monthly),
+      PayPalPaymentFields(validBaid)
+    ))
```
johnduffell: I changed the tests to assert everything in one go so that we don't get failures just saying "true was not false" or vice versa

src/main/scala/com/gu/support/workers/GetRecurringSubscription.scala
rupertbates: I have to say I'm with Jonny, I feel like scope is an odd name for this. What does scope have to do with determining whether a user has already made a contribution or not?

johnduffell: yes agreed I'm not sure if you can see the latest commit but I have changed it to say `isAccountAccessAllowed` @rupertbates 
There is still one thing called AccessScope but that's in support-models and I was too lazy to rename and do a PR there.  Happy to do so if people would like though.

build.sbt
```diff
@@ -43,7 +43,7 @@ lazy val root = (project in file("."))
       "ch.qos.logback" % "logback-classic" % "1.2.3",
       "io.symphonia" % "lambda-logging" % "1.0.1",
       "com.gu" %% "support-internationalisation" % "0.9",
-      "com.gu" %% "support-models" % "0.32",
+      "com.gu" %% "support-models" % "0.35-SNAPSHOT",
```
johnduffell: https://github.com/guardian/support-models/pull/44

src/main/scala/com/gu/zuora/model/response/Responses.scala
```diff
@@ -32,7 +32,7 @@ object ZuoraErrorResponse {
 }
 
 case class ZuoraErrorResponse(success: Boolean, errors: List[ZuoraError])
-  extends Throwable(errors.asJson.spaces2) with ZuoraResponse {
+    extends Throwable(errors.asJson.spaces2) with ZuoraResponse {
```
johnduffell: I've no idea why this file gets edited by using `sbt compile`! it's really annoying

rupertbates: It's Scalariform messing with your head!

src/test/scala/com/gu/support/workers/AccessScopeCodecTest.scala
```diff
@@ -1,19 +1,19 @@
...
-import com.gu.support.workers.model.AccessScope
-import com.gu.support.workers.model.AccessScope._
+import com.gu.support.workers.model.AccountAccessScope
+import com.gu.support.workers.model.AccountAccessScope._
 import org.scalatest.{FlatSpec, Matchers}
 
-class AccessScopeCodecTest extends FlatSpec with Matchers {
+class AccountAccessScopeCodecTest extends FlatSpec with Matchers {
 
   it should "for backwards compatibility, deserialise when it's not specified to no restriction" in {
```
rupertbates: Change the name from '... no restriction' to '... authenticated'

johnduffell: oops only just saw these, good spot!
I feel like test descriptions are like comments - they are naturally not kept up to date, they are not telling us anything that we shouldn't be able to get from the code, and they can easily get attached to the wrong thing

johnduffell: https://github.com/guardian/support-workers/pull/150

PR #148, opened 2018-10-25
faster cpu to improve cold start time

johnduffell:
## Why are you doing this?

The lambda cold start time seems very slow.  The CPU allocation is related to the memory requested, so even though we need hardly any memory, we could increase the memory allocation in order to improve the cold starts.

I have increased from 0.5 to 1.5 gb because this is what the new product api has and it seems a lot quicker https://github.com/guardian/zuora-auto-cancel/blob/master/handlers/new-product-api/cfn.yaml#L122

### DIFF
--- a/cloud-formation/src/templates/lambda.yaml
+++ b/cloud-formation/src/templates/lambda.yaml
@@ -5,7 +5,7 @@
     {{> environmentVariables}}
     FunctionName: !Sub ${Stack}-{{name}}Lambda-${Stage}
     Handler: "com.gu.support.workers.lambdas.{{name}}::handleRequest"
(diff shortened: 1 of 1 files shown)


jacobwinch:
FYI this project is set up to build and automatically update CFN (via TeamCity and RiffRaff), so you should be good to go!

johnduffell:
thanks for the hint Jacob that was helpful

I have tested in CODE and the previous cold start took 75.5 seconds, and this one took 37.5 seconds.  So still not very good, but still a doubling in speed.

Our total AWS lambda compute.memory bill from last month for all lambdas in the account was a grand total of $18.87 so even if it were all step functions and it didn't speed them up at all that would be an extra 36$

PR #149, opened 2018-10-25
Notify contributions team about support-workers errors

jacobwinch:
## Why are you doing this?

I noticed that these alerts only go to S&C, but the Contributions team are also working on this app.

I'm assuming this is the most appropriate topic, based on: https://github.com/guardian/payment-api/blob/master/src/main/resources/cloud-formation.yaml#L529-L530


johnduffell:
good idea, thanks 👍 🚨

PR #150, opened 2018-10-26
make the test names and descriptions sensible

johnduffell:
## Why are you doing this?

@rupertbates [pointed out](https://github.com/guardian/support-workers/pull/147#pullrequestreview-168026962) that in my haste to fix the names of everything, I didn't check over the tests properly.  This PR sorts that out hopefully.


src/test/scala/com/gu/support/workers/IsAccountAccessAllowedSpec.scala
```diff
@@ -4,13 +4,13 @@ import com.gu.support.workers.model.AccountAccessScope.{SessionAccess, Authentic
 import com.gu.zuora.GetAccountForIdentity.{CreatedInSession, NotCreatedInSession}
 import org.scalatest.{FlatSpec, Matchers}
 
-class GetRecurringSubscriptionIsInAccountAccessScopeSpec extends FlatSpec with Matchers {
+class IsAccountAccessAllowedSpec extends FlatSpec with Matchers {
```
johnduffell: given that half the idea of extracting the function was to make the test more reasonable, I should have remembered to update the test name to match!

PR #156, opened 2018-11-19
Fix integration tests

rupertbates:
## Why are you doing this?

A few of the integration tests had become broken due to recent changes, this PR fixes those tests or removes them where they are no longer needed. I have also added a pre-push git hook to make sure that the tests stay green in future.

[**Trello Card**](https://trello.com/c/sYEQ9kbt/2084-post-deploy-test-for-landing-pages)

### DIFF
--- a/scripts/pre-push
+++ b/scripts/pre-push
@@ -1,0 +1,10 @@
+echo "Running integration tests before push"
+sbt it:test
+rc=$?
+echo "Result" + $rc
+if [[ $rc != 0 ]]; then
+  echo "There were failures in the integration tests, please fix before pushing"
+else
+  echo "Integration tests passed successfully"
+fi
+exit $rc;
--- a/src/main/scala/com/gu/ophan/AcquisitionService.scala
+++ b/src/main/scala/com/gu/ophan/AcquisitionService.scala
@@ -2,7 +2,7 @@ package com.gu.ophan
 
 import com.gu.okhttp.RequestRunners
 
-object OphanService {
+object AcquisitionService {
 
   def apply(isTestService: Boolean): com.gu.acquisition.services.AcquisitionService =
     if (isTestService) {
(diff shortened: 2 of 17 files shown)


scripts/pre-push
```diff
@@ -0,0 +1,10 @@
+echo "Running integration tests before push"
+sbt it:test
```
johnduffell: not a bad idea!!

src/main/scala/com/gu/ophan/AcquisitionService.scala
```diff
@@ -2,7 +2,7 @@ package com.gu.ophan
 
 import com.gu.okhttp.RequestRunners
 
-object OphanService {
+object AcquisitionService {
```
johnduffell: yes that did confuse me a bit before as well, good improvement

src/main/scala/com/gu/zuora/model/response/Responses.scala
```diff
@@ -77,20 +77,13 @@ object SubscribeResponseAccount {
...
-  totalTcv: Float,
-  subscriptionId: String,
-  totalMrr: Float,
-  paymentTransactionNumber: String,
-  accountId: String,
-  gatewayResponseCode: String,
-  invoiceNumber: String,
-  invoiceId: String,
-  success: Boolean
+    accountNumber: String,
```
johnduffell: this always happens to me as well for some reason, but it was actually correct before!

johnduffell:
looks great, thanks for going through all this and nice idea making it a pre push.  👍

src/main/scala/com/gu/zuora/model/response/Responses.scala
rupertbates: It's Scalariform messing about, I've reverted this change

PR #157, opened 2018-11-20
Store request id instead of session id for zuora deduping

johnduffell:
## Why are you doing this?
from https://github.com/guardian/support-models/pull/48
Originally we added deduplication on identity id to prevent multiple contributions in a very watertight way. This was because sometimes we saw duplicates when the step functions retried the create subscription step.
We added the ability to contribute when not logged in, meaning that in a previous PR I added session id so we would deduplicate all requests that came from the same browser when logged in, but didn't deduplicate if the person wasn't logged in. This would prevent as best as possible anyone who ended up double paying without leaking any information.
Unfortunately it turned out that we can only actually deduplicate in that way within a single execution of the step functions, as currently once the zuora step passes, the analytics events are sent regardless.

[**Trello Card**](https://trello.com/c/k0RXPuuf/179-dups-change-step-functions-to-only-send-events-if-a-contribution-happened)

## Changes
This PR changes support workers to use CreatedRequestId__c in zuora to deduplicate.  This means that if the person runs the create twice from the same browser, they will end up with two subscriptions and two analytics events.


rupertbates:
"Unfortunately it turned out that we can only actually deduplicate in that way within a single execution of the step functions, as currently once the zuora step passes, the analytics events are sent regardless." Can you explain this in a bit more detail?

johnduffell:
> Unfortunately it turned out that we can only actually deduplicate in that way within a single execution of the step functions, as currently once the zuora step passes, the analytics events are sent regardless.

@rupertbates basically in the current step function structure,  once the zuora step passes, the analytics will always be sent.  This means that if there is an existing contribution previously, the zuora step will pass even though it didn't create a subscription.  Then the analytics will be incorrect because they claim an acquisition we never got.
To keep the one to one mapping between created subs and analytics events, we must only pass the zuora step if we actally create a sub in this particular step function execution.  This would be based on an id from this specific execution, rather than a global one.

johnduffell:
@rupertbates additionally we could also deduplicate on the existing logic (session id/identity id) and just go to the checkout failure state if we detected a duplicate that already exists. that would not be hard, but we will check that on the client side anyway so it should not be common.
We can never be watertight anyway for non logged in users due to the data protection reasons, so we will always have to deal with duplicates somewhere, so I'm not sure if it's worth carrying the complexity to reduce the chance of it happening.

johnduffell:
seems to work fine in CODE (although I did notice a previous execution that was looping on create SF contact since 2am, so I terminated it)

PR #159, opened 2018-12-06
Promo codes

rupertbates:
## Why are you doing this?
This PR allows us to use promo codes with the Digital Pack sign up.

## Changes
* The vast majority of changes in this PR are a result of moving the Zuora model classes into the support-models library so that they are accessible to the new support-services library which provides promo code related functionality to support-frontend and support-workers.
* I have removed some code and tests which were related to handling an old version of the input schema from support-models - this is no longer necessary as support-frontend has been using the new schema exclusively for months.

I will put comments in the PR to highlight areas where there have been actual functional changes

[**Trello Card**](https://trello.com/c/fEWP5zGf/2059-promo-code-support-for-buying-digital-packs-via-support-workers)


src/main/scala/com/gu/services/Services.scala
```diff
@@ -25,5 +26,6 @@ class Services(isTestUser: Boolean) {
   lazy val salesforceService = new SalesforceService(salesforceConfigProvider.get(isTestUser), configurableFutureRunner(40.seconds))
   lazy val zuoraService = new ZuoraService(zuoraConfigProvider.get(isTestUser), configurableFutureRunner(60.seconds))
   lazy val acquisitionService = AcquisitionService(isTestUser)
+  lazy val promotionService = new PromotionService(promotionsConfigProvider.get(isTestUser))
```
rupertbates: New service to deal with promotions

src/main/scala/com/gu/config/Configuration.scala
```diff
@@ -29,6 +28,7 @@ object Configuration {
   val payPalConfigProvider = new PayPalConfigProvider(config, stage)
   val salesforceConfigProvider = new SalesforceConfigProvider(config, stage)
   val zuoraConfigProvider = new ZuoraConfigProvider(config, stage)
+  val promotionsConfigProvider = new PromotionsConfigProvider(config, stage)
```
rupertbates: New config for promotions

src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -69,61 +67,25 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
-        val contractAcceptanceDate = contractEffectiveDate
-          .plusDays(config.digitalPack.defaultFreeTrialPeriod)
-          .plusDays(config.digitalPack.paymentGracePeriod)
-
-        buildProductSubscription(
-          config.digitalPackRatePlan(d.billingPeriod),
-          contractAcceptanceDate = contractAcceptanceDate,
-          contractEffectiveDate = contractEffectiveDate
-        )
+      case c: Contribution => c.build(config)
```
rupertbates: Extracted to a separate class `ProductSubscriptionBuilders` as it was getting quite complex

src/main/scala/com/gu/zuora/ProductSubscriptionBuilders.scala
```diff
@@ -0,0 +1,68 @@
...
+
+      val productRatePlanId = config.digitalPackRatePlan(digitalPack.billingPeriod)
+
+      val subscriptionData = buildProductSubscription(
+        productRatePlanId,
+        contractAcceptanceDate = contractAcceptanceDate,
+        contractEffectiveDate = contractEffectiveDate
+      )
+
+      maybePromoCode
```
rupertbates: This is where promo codes are applied - the `PromotionService` updates the `subscriptionData` object to apply either a discount rate plan (for discounts) or to extend the contract acceptance date (for free trials). If the code is not applicable it will simply return the original `subscriptionData` unmodified. 

The `isRenewal` parameter is hard coded to false for now because we only deal with new subscriptions in support-workers. If this changes in future then we will need to pass in the correct value here.

src/test/resources/logback.xml
```diff
@@ -0,0 +1,15 @@
+<configuration>
```
rupertbates: Stop Sentry from firing during tests

src/test/scala/com/gu/support/workers/lambdas/FailureHandlerSpec.scala
```diff
@@ -50,22 +50,6 @@ class FailureHandlerSpec extends LambdaSpec {
 
   }
 
-  "FailureHandler lambda" should "still return a failed JsonWrapper and an Unknown failure reason if it receives the old schema" in {
```
rupertbates: We will never be receiving the old schema any more so I've removed code and tests that refer to it

PR #161, opened 2018-12-13
Use abTests in ReferrerAcquisitionData

joelochlann:
Supposedly, `abTest` is deprecated and `abTests` is the one we should be using (according to [this comment](https://github.com/guardian/acquisition-event-producer/blob/91ff314c2f22eac82f1e7594cee6a300e6681873/src/main/scala/com/gu/acquisition/model/ReferrerAcquisitionData.scala#L19)).

@Ap0c followed this advice in https://github.com/guardian/support-frontend/pull/997 when he changed the clientside model from `abTest: ?AcquisitionABTest` to `abTests: ?AcquisitionABTest[]`. But unfortunately there's a bit in the middle here in support-workers which was only looking at `abTest`.

This is why we haven't been getting A/B test data for recurring contributions for the last few months. (We've had component ids as a workaround, which is why we didn't prioritise fixing it til now)

@Mullefa can you remember what the story is with that deprecation? It would be nice to remove the deprecated `abTest` field everywhere, but I seem to remember that Thrift's position-sensitive data encoding strategy makes that problematic.

### DIFF
--- a/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
+++ b/src/main/scala/com/gu/support/workers/lambdas/SendAcquisitionEvent.scala
@@ -101,7 +101,11 @@ object SendAcquisitionEvent {
               // Currently only passing through at most one campaign code
               campaignCode = data.referrerAcquisitionData.campaignCode.map(Set(_)),
               abTests = Some(thrift.AbTestInfo(
-                data.supportAbTests ++ Set(data.referrerAcquisitionData.abTest).flatten
+                data.supportAbTests
+                  // abTest is deprecated and support-frontend now sends abTests.
+                  // but let's add abTest as well just in case.
+                  ++ data.referrerAcquisitionData.abTests.getOrElse(Set())
+                  ++ Set(data.referrerAcquisitionData.abTest).flatten
               )),
               countryCode = Some(state.user.country.alpha2),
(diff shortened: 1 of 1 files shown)


johnduffell:
great work finding that! 🎉 👍   - although if `abTest` is probably not used, it sounds like some tech debt we should just remove completely?

joelochlann:
yep, see above

> @Mullefa can you remember what the story is with that deprecation? It would be nice to remove the deprecated abTest field everywhere, but I seem to remember that Thrift's position-sensitive data encoding strategy makes that problematic.

johnduffell:
oh fair enough didn't realise it was thrift, I am not a thrift expert!

Mullefa:
@joelochlann - I _think_ when the data was initially modelled we were just considering the test the reader would come from on the main website - not e.g. support frontend. It might be possible to remove the `abTest` field from the thrift model - not sure 😕

joelochlann:
I think I might actually be talking rubbish - that model does use bits from the thrift definition but it just lives inside `acquisitions-event-producer`. So it seems totally possible to remove it. Just need to do it in the library and then bump versions across projects. There may be some tweaks needed to those projects if they're using the deleted field current but should be trivial.

Here's the `acquisitions-event-producer` PR: https://github.com/guardian/acquisition-event-producer/pull/43

PR #162, opened 2018-12-17
Send acquisition events to kinesis

tomrf1:
## Why are you doing this?
Adds a third acquisition service (in addition to ophan and GA) - a kinesis stream.
This will initially be used for the contributions ticker calculation.

Note - a credentials provider is not required by the kinesis client if running in a lambda. But it is required in order to run the integration tests locally.

### DIFF
--- a/cloud-formation/src/templates/cfn-template.yaml
+++ b/cloud-formation/src/templates/cfn-template.yaml
@@ -16,6 +16,9 @@ Parameters:
   OphanRole:
     Type: String
     Description: ARN of the Ophan cross-account role
+  KinesisStreamArn:
+    Type: String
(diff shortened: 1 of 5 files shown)


cloud-formation/src/templates/cfn-template.yaml
```diff
@@ -16,6 +16,9 @@ Parameters:
   OphanRole:
     Type: String
     Description: ARN of the Ophan cross-account role
+  KinesisStreamArn:
+    Type: String
+    Description: ARN of the kinesis stream to write events to
```
johnduffell: is that "acquisition events" or also other types of events?

tomrf1: just acquisition

PR #168, opened 2019-01-15
Preview upcoming payments and pass information through to thank you email lambda

jacobwinch:
## Why are you doing this?

This PR allows us to preview a user's payment schedule before creating a new subscription. This will allow us to provide accurate billing information in thank you emails (for subscriptions products), which becomes more important in complex scenarios where promotions and discounts are involved. 

Note that the email updates which make use of this payment schedule will be added in a subsequent PR.

## Changes

* Use latest version of support-models and pass payment schedule through as part of SendThankYouEmailState
* Add code for previewing invoices via Zuora's REST API
* Refactor CreateZuoraSubscription slightly to accommodate preview
* Add (and update) unit tests

### DIFF
--- a/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
+++ b/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
@@ -34,17 +34,23 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
       maybeDomainSubscription <- GetRecurringSubscription(services.zuoraService, state.requestId, identityId, state.product.billingPeriod)
       previewPaymentSchedule <- PreviewPaymentSchedule(subscribeItem, state.product.billingPeriod, services, checkSingleResponse)
       thankYouState <- maybeDomainSubscription match {
-        case Some(domainSubscription) => skipSubscribe(state, requestInfo, domainSubscription)
+        case Some(domainSubscription) => skipSubscribe(state, requestInfo, previewPaymentSchedule, domainSubscription)
         case None => subscribe(state, subscribeItem, previewPaymentSchedule, requestInfo, services)
       }
     } yield thankYouState
   }
 
-  def skipSubscribe(state: CreateZuoraSubscriptionState, requestInfo: RequestInfo, subscription: DomainSubscription): FutureHandlerResult = {
+  def skipSubscribe(
+    state: CreateZuoraSubscriptionState,
+    requestInfo: RequestInfo,
+    previewedPaymentSchedule: PaymentSchedule,
+    subscription: DomainSubscription
+  ): FutureHandlerResult = {
     val message = "Skipping subscribe for user because they are already an active contributor"
     SafeLogger.info(message)
     FutureHandlerResult(
-      getEmailState(state, subscription.accountNumber, subscription.subscriptionNumber, PaymentSchedule(Nil)), requestInfo.appendMessage(message)
+      getEmailState(state, subscription.accountNumber, subscription.subscriptionNumber, previewedPaymentSchedule),
+      requestInfo.appendMessage(message)
     )
   }
 
--- a/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
+++ b/src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
@@ -9,7 +9,7 @@ import com.gu.support.promotions.PromotionService
 import com.gu.support.workers.states.{CreateZuoraSubscriptionState, SendThankYouEmailState}
 import com.gu.support.workers._
 import com.gu.support.zuora.api._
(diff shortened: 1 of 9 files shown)


src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -27,36 +27,57 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
-      case Some(domainSubscription) => skipSubscribe(state, requestInfo, domainSubscription)
-      case None => subscribe(state, requestInfo, services)
-    }
-  } yield thankYouState
+  ): FutureHandlerResult = {
+    val subscribeItem = buildSubscribeItem(state, services.promotionService)
+    for {
+      identityId <- Future.fromTry(IdentityId(state.user.id))
+      maybeDomainSubscription <- GetRecurringSubscription(services.zuoraService, state.requestId, identityId, state.product.billingPeriod)
+      previewPaymentSchedule <- PreviewPaymentSchedule(subscribeItem, state.product.billingPeriod, services, checkSingleResponse)
```
jacobwinch: The preview call is made in the CreateZuoraSubscription lambda. I've put this here because we could use the result to perform safety checks before actually setting up subscriptions in the future.

lmath: Good idea 👍

src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -27,36 +27,57 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
+    state: CreateZuoraSubscriptionState,
+    subscribeItem: SubscribeItem,
+    previewedPaymentSchedule: PaymentSchedule,
+    requestInfo: RequestInfo,
+    services: Services
+  ): FutureHandlerResult =
+    singleSubscribe(services.zuoraService.subscribe)(subscribeItem)
       .map(response =>
-        HandlerResult(getEmailState(state, response.domainAccountNumber, response.domainSubscriptionNumber), requestInfo))
+        HandlerResult(getEmailState(state, response.domainAccountNumber, response.domainSubscriptionNumber, previewedPaymentSchedule), requestInfo))
```
johnduffell: it doesn't seem necessary to pass this in only to pass it straight on to the HandlerResult.  This also applies to requestInfo which is existing.

src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
```diff
@@ -27,36 +27,57 @@ class CreateZuoraSubscription(servicesProvider: ServiceProvider = ServiceProvide
...
+      }
+    } yield thankYouState
+  }
 
   def skipSubscribe(state: CreateZuoraSubscriptionState, requestInfo: RequestInfo, subscription: DomainSubscription): FutureHandlerResult = {
     val message = "Skipping subscribe for user because they are already an active contributor"
     SafeLogger.info(message)
-    FutureHandlerResult(getEmailState(state, subscription.accountNumber, subscription.subscriptionNumber), requestInfo.appendMessage(message))
+    FutureHandlerResult(
+      getEmailState(state, subscription.accountNumber, subscription.subscriptionNumber, PaymentSchedule(Nil)), requestInfo.appendMessage(message)
```
johnduffell: so if the lambda retries we will lose the payment schedule.  is that alright?  I don't think the lambda should returns something different if it retries.

johnduffell:
great good testing as well, see my comments but the main one is can we really accept an empty billing preview if the lambda retries for some reason?

src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
jacobwinch: Very good point, I'll update this.

src/main/scala/com/gu/support/workers/lambdas/CreateZuoraSubscription.scala
jacobwinch: We could do something like this instead:

```
thankYouState <- maybeDomainSubscription match {
  case Some(domainSubscription) => skipSubscribe(state, requestInfo, previewPaymentSchedule, domainSubscription)
  case None => subscribe(state, subscribeItem, services).map(response => toHandlerResult(state, response, previewPaymentSchedule, requestInfo))
}
```

Do you think that's preferable?

johnduffell: yes that sounds better, then we reduced the number of parameters from 5 to 3.  Also your way looks more normal that what I imagined in my head! 💯

PR #170, opened 2019-01-23
If we have a telephone number when creating a salesforce contact, add it too.

lmath:
## Why are you doing this?
We are asking for telephone number in the new digital pack checkout. When we make a request to create a new digital pack, if the user gives us their telephone number, we will include it in that request. We should make sure that telephone number is saved in salesforce, so that we can use it to contact them regarding their subscription if relevant. 

[**Trello Card**](https://trello.com/c/zcaNKssy/2156-make-sure-telephone-number-if-provided-makes-its-way-to-salesforce)

## Changes

* Use new version of support-models https://github.com/guardian/support-libraries/pull/16
* When we create a new salesforce contact, if we have a telephone number for the user, then we add this too the contact too.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -52,7 +52,7 @@ lazy val root = (project in file("."))
       "ch.qos.logback" % "logback-classic" % "1.2.3",
       "io.symphonia" % "lambda-logging" % "1.0.1",
       "com.gu" %% "support-internationalisation" % "0.9",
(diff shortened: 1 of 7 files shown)


lmath:
Thanks @jacobwinch . Some phone number validation is a good idea. I will add it in when I raise my support-frontend PR. And thanks for the salesforce link; I was looking everywhere for that!

PR #171, opened 2019-01-31
Testing promotions

rupertbates:
## Why are you doing this?

[**Trello Card**](https://trello.com)

## Changes

* Change 1
* Change 2


src/test/scala/com/gu/support/workers/errors/PayPalErrorsSpec.scala
```diff
@@ -10,10 +10,12 @@ import com.gu.support.workers.Fixtures.{createPayPalPaymentMethodDigitalPackJson
 import com.gu.support.workers.LambdaSpec
 import com.gu.support.workers.exceptions.RetryUnlimited
 import com.gu.support.workers.lambdas.CreatePaymentMethod
+import com.gu.test.tags.annotations.IntegrationTest
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
 
+@IntegrationTest
```
jacobwinch: Sorry if I'm being slow, but why does this need to be tagged as a result of this PR?

rupertbates: I'm not sure it does actually. Don't know where that came from!

PR #173, opened 2019-02-13
Save address details in billing address (not mailing address) in SF

lmath:
## Why are you doing this?
`support-workers` currently support recurring contributions and digital subscriptions, neither of which actually require mailing anything. 

Additionally, the digital pack checkout on subscribe.theguardian saves address details to billing address. 

## Changes

* When creating a salesforce contact, send country and (optional) state to billing address rather than mailing address (note that this entails sending them to OtherAddress à la https://github.com/guardian/membership-common/blob/master/src/main/scala/com/gu/salesforce/ContactDeserializer.scala)


lmath:
@paulbrown1982 - Can you think of any complications this might cause?

PR #176, opened 2019-02-15
Use billing address to get fields to GoCardless and Salesforce

lmath:
## Why are you doing this?
We will now send billing address through from `support-frontend` as of https://github.com/guardian/support-frontend/pull/1514 and I've moved the responsibility to convert addresses collected from the form -> addresses required by Zuora (eventually GoCardless via the Zuora-GoCardless integration) or Salesforce. 

So, this PR uses the billing address fields to:
1. send along to Zuora (eventually GC)
2. send to the billing address in Salesforce

More info here: https://github.com/guardian/support-libraries/pull/23

## Changes

* Use latest version of `support-services` (know to receive billing address)
* Continue to ensure that address info is sent on the request to Zuora when the user has paid with direct debit, but get this info from the billing address
* If we receive address info, send it to the billing address in Salesforce


src/main/scala/com/gu/support/workers/lambdas/CreateSalesforceContact.scala
```diff
@@ -16,13 +18,18 @@ class CreateSalesforceContact extends ServicesHandler[CreateSalesforceContactSta
 
   override protected def servicesHandler(state: CreateSalesforceContactState, requestInfo: RequestInfo, context: Context, services: Services) = {
     SafeLogger.debug(s"CreateSalesforceContact state: $state")
+
+    val addressLine = combinedAddressLine(state.user.billingAddress.lineOne, state.user.billingAddress.lineTwo) map asFormattedString
```
rupertbates: I think everywhere else in the codebase we would write `.map(asFormattedString)` might be worth keeping it consistent

lmath: Sure. Will do.

PR #177, opened 2019-02-15
Support-services to 0.10

lmath:
## Why are you doing this?
https://github.com/guardian/support-libraries/pull/25 removed some fields that we no longer need now that address info is available via the billingAddress associated with the user. So, we use version 0.10 and make sure to get country and state from billing address.

## Changes

* Bump support-services to 0.10


build.sbt
```diff
@@ -51,6 +51,7 @@ lazy val root = (project in file("."))
       "ch.qos.logback" % "logback-classic" % "1.2.3",
       "io.symphonia" % "lambda-logging" % "1.0.1",
       "com.gu" %% "support-services" % "0.9",
+      "com.gu" %% "support-models" % "0.59",
```
lmath: I'm waiting for the support-services 0.10 artefact to appear

PR #180, opened 2019-02-21
When creating digital pack subscriptions, get rate plan id from the Product mapping

lmath:
## Why are you doing this?
Now that there is a mapping that provides RatePlanIds here:
https://github.com/guardian/support-libraries/blob/master/support-models/src/main/scala/com/gu/support/catalog/Product.scala

We should get the RatePlanIds for digital pack from that mapping instead, and then it doesn't need to be maintained in `support-config`.

Then we can also do the same for print products. 

[**Trello Card**](https://trello.com/c/ofdmiV3O/2253-add-paper-product-types-to-support-models-config-to-support-config)

## Changes

* Get mapping of ratePlanIds from Product when a ratePlanId is needed to create a digital pack

### DIFF
--- a/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
+++ b/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
@@ -3,6 +3,7 @@ package com.gu.support.workers.exceptions
 import java.net.{SocketException, SocketTimeoutException}
 
 import com.amazonaws.services.kms.model._
+import com.amazonaws.services.s3.model.AmazonS3Exception
 import com.amazonaws.services.sqs.model.{AmazonSQSException, InvalidMessageContentsException, QueueDoesNotExistException}
 import com.gu.acquisition.model.errors.AnalyticsServiceError
 import com.gu.helpers.WebServiceHelperError
@@ -41,6 +42,13 @@ object RetryImplicits {
     def asRetryException: RetryException = throwable match {
       case e: InvalidMessageContentsException => new RetryNone(message = e.getMessage, cause = throwable)
       case e: QueueDoesNotExistException => new RetryLimited(message = e.getMessage, cause = throwable)
+    }
+  }
+
+  implicit class ZuoraCatalogueConversions(val throwable: ZuoraCatalogException) {
+    def asRetryException: RetryException = throwable match {
+      case e: CatalogDataNotFound => new RetryNone(message = e.getMessage, cause = throwable)
+      case e: ZuoraCatalogException => new RetryNone(message = e.getMessage, cause = throwable)
     }
   }
 
--- a/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
+++ b/src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
@@ -45,7 +45,7 @@ object RetryImplicits {
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/main/scala/com/gu/support/workers/exceptions/RetryImplicits.scala
```diff
@@ -44,6 +45,10 @@ object RetryImplicits {
     }
   }
 
+  implicit class ZuoraCatalogueConversions(val throwable: CatalogDataNotFoundException) {
```
walaura: nit: could we have consistent spelling between catalog/catalogue?

lmath: Yep. 👍 I corrected back to catalog a few times but missed this one

src/main/scala/com/gu/zuora/ProductSubscriptionBuilders.scala
```diff
@@ -1,15 +1,34 @@
...
 
+  def getProductRatePlanId[PT <: ProductType, P <: Product](product: P, productType: PT): ProductRatePlanId = {
+    val touchpointEnvironment = TouchPointEnvironments.fromStage(Configuration.stage)
+    def getRatePlans[T <: Product](product: T): Seq[ProductRatePlan[Product]] = product.ratePlans.getOrElse(touchpointEnvironment, Nil)
+
+    val ratePlans: Seq[ProductRatePlan[Product]] = getRatePlans(catalog.DigitalPack)
+
+    val maybeProductRatePlanId: Option[ProductRatePlanId] = ratePlans.find(p => p.billingPeriod == productType.billingPeriod) map (_.id)
+
+    Try(maybeProductRatePlanId.get) match {
```
walaura: Is there a reasons this needs a `try`?

lmath: I want to move away from keeping `ProductRatePlanId`s among Zuora config here: https://github.com/guardian/support-libraries/blob/master/support-config/src/main/scala/com/gu/support/config/ZuoraConfigProvider.scala 

We're already maintaining a mapping as [here](https://github.com/guardian/support-libraries/blob/master/support-models/src/main/scala/com/gu/support/catalog/Product.scala), and I think these ids belong to a notion of the Zuora catalog rather than Zuora related config anyway. 

But the upside of keeping these ids among Zuora config is they were required config as soon as we tried to load the config, if there was a `ProductRatePlanId` missing that we needed this would fail fast. If we make a mistake and we're missing a `ProductRatePlanId` that should be defined, then when we try to `maybeProductRatePlanId` on line 24 will be a None. Then, `maybeProductRatePlanId.get` will throw `java.util.NoSuchElementException: None.get`. So, I wrapped it in a custom error so that if that happens we'll get a more descriptive error in the logs. 

But I'm not 100% sure this is the best way to handle this so v v open to ideas!

lmath: Maybe you have some thoughts on this @jacobwinch ?

jacobwinch: I like the custom exception - it means I'd be able to understand a failure simply by looking at a failed execution in AWS Console without any further digging 💯👍

PR #181, opened 2019-02-22
Revert "When creating digital pack subscriptions, get rate plan id from the Product mapping"

lmath:
Reverts guardian/support-workers#180


lmath:
I was testing with a UAT user in PROD, and it turns out that I forgot to add support for test users in https://github.com/guardian/support-workers/pull/180 

This caused the subscribe request to go to UAT zuora with the PROD ProductRatePlanId, and that caused the failure. 

This wouldn't have affected PROD users. I will raise https://github.com/guardian/support-workers/pull/180  again (with test user support) shortly.

PR #182, opened 2019-02-22
Get rate plan id from product (again) 

lmath:
## Why are you doing this?
I reverted https://github.com/guardian/support-workers/pull/180 when I tried going through checkout with a test user in prod and then the corresponding state machine execution failed. 

Turns out this is because I needed to add that when `isTestUser` is true, I need to get ProductRatePlanIds for UAT, otherwise from the default stage. So this PR is https://github.com/guardian/support-workers/pull/180 with that added!


src/main/scala/com/gu/zuora/ProductSubscriptionBuilders.scala
```diff
@@ -15,8 +17,9 @@ import scala.util.{Failure, Success, Try}
 
 object ProductSubscriptionBuilders {
 
-  def getProductRatePlanId[PT <: ProductType, P <: Product](product: P, productType: PT): ProductRatePlanId = {
-    val touchpointEnvironment = TouchPointEnvironments.fromStage(Configuration.stage)
+  def getProductRatePlanId[PT <: ProductType, P <: Product](product: P, productType: PT, isTestUser: Boolean): ProductRatePlanId = {
+    val touchpointEnvironment = if (isTestUser) UAT else TouchPointEnvironments.fromStage(Configuration.stage)
```
jacobwinch: Sorry I should have spotted this during code review last time but it had been a long day!

